PR #26, opened 2024-02-02
Fix arguments ignored if no config file found

awoimbee:
[sc-16887]
When using `from_config` and no config file is found, but kwargs were passed, we should not throw them away.


tmpbeing (approved):
Thanks

PR #29, opened 2024-02-09
fix: Double description of client config vars

tmpbeing:


awoimbee:
I think the class level docs is for sphinx and the field level is for clients that want to extend the SDK (eg if someone wanted to write a cli c; ). This comes from PR 961 on the _other_ repo.
If you've checked that the docs are not impacted let's merge, otherwise I think we're stuck with duplicated field docs.

tmpbeing:
It does in fact work as expected

PR #30, opened 2024-02-12
Direction for Windows-developer mode

yias:
[sc-16695]

This PR adds install directions on Windows for the developer track

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -69,21 +69,51 @@ need to follow these steps:
 
       git clone https://github.com/ansys/pysimai
 
-#. `Install PDM <https://pdm-project.org/latest/#recommended-installation-method>`_
+#. `Install PDM <https://pdm-project.org/latest/#recommended-installation-method>`_. NB: If you are a Windows user, make sure that Python is installed on your system and it is added to the Path.
 
 #. Use PDM to create a virtual env with the dependencies and activate it
 
-   .. code:: bash
+   * For Linux/Mac:
+      .. code:: bash
 
-      pdm install
-      eval $(pdm venv activate)
+         pdm install
+         eval $(pdm venv activate)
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -69,21 +69,51 @@ need to follow these steps:
...
+
+         .. code:: shell
+
+            python -m pip install tox
+
+      * and then run tox
+
+         .. code:: shell
+
+            tox
```
tmpbeing, line 115:
```suggestion
   .. code:: bash

         python -m pip install tox
         tox
```

Trying to reduce noise by assuming the user has pip by that point and these commands work for both systems. If the user doesn't have pip I'm sure the error message and google will be enough.

README.rst
```diff
@@ -69,21 +69,51 @@ need to follow these steps:
 
       git clone https://github.com/ansys/pysimai
 
-#. `Install PDM <https://pdm-project.org/latest/#recommended-installation-method>`_
+#. `Install PDM <https://pdm-project.org/latest/#recommended-installation-method>`_. NB: If you are a Windows user, make sure that Python is installed on your system and it is added to the Path.
```
tmpbeing, line 72:
```suggestion
#. `Install PDM <https://pdm-project.org/latest/#recommended-installation-method>`_.
```

I think we should assume developers have python properly installed on their system.

tmpbeing (approved):
Thanks. Made some suggestion to try and keep it less verbose.
Eventually we'll have to move some things to a contribute section of the doc.

README.rst
awoimbee: Just saw this PR.
So if I'm not a windows user I don't need to have "Python is installed on your system and it is added to the Path" ?

PR #32, opened 2024-02-16
fix monitor_callback interface not respected in upload_file_with_presigned_post

awoimbee:
This function gave the whole `MultipartEncoderMonitor` to `monitor_callback` instead of just an `int`

### DIFF
--- a/src/ansys/simai/core/api/mixin.py
+++ b/src/ansys/simai/core/api/mixin.py
@@ -201,22 +201,19 @@ class ApiClientMixin:
         filename = getattr(file, "name", "")
         upload_form["file"] = (filename, file, "application/octet-stream")
         multipart = MultipartEncoder(upload_form)
-
         if monitor_callback is not None:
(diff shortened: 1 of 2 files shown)


src/ansys/simai/core/api/mixin.py
```diff
@@ -201,22 +201,19 @@ def upload_file_with_presigned_post(
         filename = getattr(file, "name", "")
         upload_form["file"] = (filename, file, "application/octet-stream")
         multipart = MultipartEncoder(upload_form)
-
         if monitor_callback is not None:
             # Wrap the monitor callback so that it receives only the bytes read
             # instead of the full MultipartEncoderMonitor object
-            def wrap_monitor_callback(monitor_callback):
-                # FIXME: This ain't gonna work chief
```
tmpbeing, line 209: 🥴

tmpbeing (approved):
Whoops, thanks

PR #36, opened 2024-04-05
Bug/sc 17799/investigate potential get subset pysimai

kliment-slice:
[sc-17799]
fix for pysimai bug on get_subset(); tested on dev1


awoimbee (approved):
lgtm

PR #37, opened 2024-04-05
Adding is_trainable and get_variables in Project

yias:
[sc-18073]

### DIFF
--- a/src/ansys/simai/core/data/projects.py
+++ b/src/ansys/simai/core/data/projects.py
@@ -81,6 +81,14 @@ class Project(DataModel):
         """Check if the project meets the prerequisites to be trained."""
         return self._client._api.is_project_trainable(self.id).status_code == 200
 
+    def get_variables(self) -> dict:
+        """Get the available variables for the model's input/output."""
+        variables = self.sample.fields.get("extracted_metadata")
+        data = {}
+        for key in variables:
+            data |= {key: [var.get("name") for var in variables.get(key).get("fields")]}
+        return data
+
 
 class ProjectDirectory(Directory[Project]):
     """Provides a collection of methods related to projects.
--- a/src/ansys/simai/core/data/projects.py
+++ b/src/ansys/simai/core/data/projects.py
@@ -77,6 +77,10 @@ class Project(DataModel):
         """Delete the project."""
         self._client._api.delete_project(self.id)
 
+    def is_trainable(self) -> bool:
+        """Check if the project meets the prerequisites to be trained."""
+        return self._client._api.is_project_trainable(self.id).status_code == 200
+
 
 class ProjectDirectory(Directory[Project]):
     """Provides a collection of methods related to projects.
--- a/src/ansys/simai/core/data/projects.py
+++ b/src/ansys/simai/core/data/projects.py
@@ -33,19 +33,28 @@ if TYPE_CHECKING:
 class IsTrainableInfo(NamedTuple):
     """Properties for project's trainability.
 
-    The objects of this class are associated with the output of the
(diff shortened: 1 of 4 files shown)


src/ansys/simai/core/data/projects.py
```diff
@@ -30,6 +30,35 @@
     from ansys.simai.core.data.training_data import TrainingData
 
 
+class IsTrainableInfo(NamedTuple):
+    """Properties for project's trainability.
+
+    The objects of this class are associated with the output of the
+    evaluation, so that it returns a boolean variable according to
+    trainability of the project:
```
awoimbee, line 38: I personally have a hard time understanding this sentence.
IMO the example + the fact that there is `def __bool__` is sufficient/more-straightforward as documentation.

src/ansys/simai/core/data/projects.py
```diff
@@ -77,6 +106,19 @@ def delete(self) -> None:
...
+        tt = self._client._api.is_project_trainable(self.id)
+        return IsTrainableInfo(**tt)
+
+    def get_variables(self) -> dict:
+        """Get the available variables for the model's input/output."""
+        variables = self.sample.fields.get("extracted_metadata")
+        data = {}
+        for key in variables:
+            data |= {key: [var.get("name") for var in variables.get(key).get("fields")]}
+        return data
```
awoimbee, line 120: 1. Seems this will crash if the project has no sample, maybe we should instead return `None`
2. Let's make the code easier to read:
```suggestion
    def get_variables(self) -> dict:
        """Get the available variables for the model's input/output."""
        sample_metadata = self.sample.fields.get("extracted_metadata")
        vars = {}
        for location, located_data in sample_metadata.items():
            local_fields = located_data.get("fields", [])
            vars[location] = [field.get("name") for field in local_fields]
```
3. IDK what your spec looks like on this one but IMO `variable` isn't a great name:
    * In the SDK everything is a variable, IMO it's confusing
    * Internally it's called a field (following VTK's naming on this one)
    * We have features planned which use `field` in the name

awoimbee (commented):
Nice features, need a bit of polish

src/ansys/simai/core/data/projects.py
yias: I updated the text, hope it's easier to understand it now

src/ansys/simai/core/data/projects.py
yias: 1. good catch! I added a check.
2. you didn't like the merge operator? 😢 I updated the code.
3. I try to follow the same naming as in the FE so it can be easier for the user to digest. The name there is `variables`
image
If you have another idea, let me know.

src/ansys/simai/core/data/projects.py
```diff
@@ -77,6 +115,23 @@ def delete(self) -> None:
         """Delete the project."""
         self._client._api.delete_project(self.id)
 
+    def is_trainable(self) -> bool:
+        """Check if the project meets the prerequisites to be trained."""
+        tt = self._client._api.is_project_trainable(self.id)
+        return IsTrainableInfo(**tt)
+
+    def get_variables(self) -> dict:
```
awoimbee, line 123:
```suggestion
    def get_variables(self) -> dict[str, list[str]] | None:
```

awoimbee (approved):
Little type issue left, otherwise LGTM !

src/ansys/simai/core/data/projects.py
yias: good point, I removed it.

src/ansys/simai/core/data/projects.py
yias: corrected

PR #38, opened 2024-04-09
Use last build configuration for a new build

yias:
## Description 

It enable launching a build using an already existing configuration from a project

Usage: 

```py
my_project = simai.projects.get("my_project")

build_conf = my_project.last_model_configuration

launched_model = simai.models.build(build_conf)

model_id = launched_model .id

model_config = launched_model .configuration
```

If the user wants to use the configuration to another project, then:

```py
my_project = simai.projects.get("my_project")

build_conf = my_project.last_model_configuration

# assuming that the other_project is trainable
other_project = simai.projects.get("other_project")

builf_conf.project_id = other_project .id

launched_model = simai.models.build(build_conf)
```

## Remarks

- Since the models are hidden (the GET request to `/<project_id>/models` returns authorization error), the method `get` of the class `ModelDirectory` raises a `NotImplementedError` error when used. Not 100% sure if this the best practice 😟 Maybe, just passing would be ok, since it is not supported yet. 

## Tests

- Tested against Dev
- Unit tests

### DIFF
--- a/src/ansys/simai/core/api/project.py
+++ b/src/ansys/simai/core/api/project.py
@@ -70,15 +70,21 @@ class ProjectClientMixin(ApiClientMixin):
     def launch_build(
         self,
         project_id: str,
-        build_config: Dict[str, Any],
-        dismiss_data_with_fields_discrepancie: bool = False,
+        config: Dict[str, Any],
+        dismiss_data_with_fields_discrepancies: bool = False,
         dismiss_data_with_volume_overflow: bool = False,
     ):
+        """Launches a build for a project and according to a given configuration.
+
+        Args:
+            project_id: the ID of the project
+            config: the build configuration
+            dismiss_data_with_fields_discrepancies: set to True for omitting training data with missing properties
+            dismiss_data_with_volume_overflow: set to True to
+
+        """
         params = {
-            "dismiss_data_with_fields_discrepancies": dismiss_data_with_fields_discrepancie,
+            "dismiss_data_with_fields_discrepancies": dismiss_data_with_fields_discrepancies,
             "dismiss_data_with_volume_overflow": dismiss_data_with_volume_overflow,
         }
-        return self._post(f"projects/{project_id}/model", json=build_config, params=params)
-
-    def get_model(self, project_id: str):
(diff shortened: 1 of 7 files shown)


src/ansys/simai/core/data/models.py
```diff
@@ -0,0 +1,84 @@
...
+    _data_model = Model
+
+    def get(self, project_id: str) -> Model:
+        """Get a model by project ID.
+
+        Args:
+            project_id: ID of the project.
+        """
+
+        return self._model_from(self._client._api.get_model(project_id))
```
yias, line 66: not sure, what it should return :/

src/ansys/simai/core/api/project.py
```diff
@@ -66,3 +66,19 @@ def set_project_sample(self, project_id: str, training_data_id: str):
 
     def delete_project(self, project_id: str):
         self._delete(f"projects/{project_id}", return_json=False)
+
+    def launch_build(
+        self,
+        project_id: str,
+        build_config: Dict[str, Any],
+        dismiss_data_with_fields_discrepancie: bool = False,
```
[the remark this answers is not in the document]
yias: yes, `Directory` has an abstract method `get` that has to have an implementation in the child classes. Would it make sense to let `pass` instead of raising an error?

src/ansys/simai/core/api/project.py
```diff
@@ -66,3 +66,19 @@ def set_project_sample(self, project_id: str, training_data_id: str):
...
+        dismiss_data_with_volume_overflow: bool = False,
+    ):
```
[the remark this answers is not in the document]
yias: documentation is added

src/ansys/simai/core/data/models.py
yias: yes, Directory has an abstract method get that has to have an implementation in the child classes. Would it make sense to let pass instead of raising an error?

tmpbeing: Assuming it doesn't block you, I'd rather have an error. Having a pass could make it harder to debug issues in the future.

yias: np

PR #39, opened 2024-04-11
Build configuration from scratch

yias:
## Description

This PR enables the definition of full build configuration from scratch. 

story: [sc-18068]

## Usage

```py
import ansys.simai.core as asc
from ansys.simai.core.data.model_configuration import (
     DomainOfAnalysis,
     ModelConfiguration,
     ModelInput,
     ModelOutput,
)

simai = asc.from_config()

my_project = simai.projects.get(name="my-project")

model_input = ModelInput(surface=["Velocity"], boundary_conditions=["Vx"])

model_output = ModelOutput(surface=["Pressure", "WallShearStress_0"], volume=["Velocity_0", "Pressure"])

global_coefficients = [("max(Pressure)", "maxpress")]

doa = DomainOfAnalysis(
    length=("relative_to_max", 5, 8.1),
    width=("relative_to_max", 5, 8.1),
    height=("absolute", -4.5, 0.1),
)

new_conf = ModelConfiguration(
     project=my_project ,
     build_preset="debug",
     continuous=False,
     input=model_input,
     output=model_output,
     global_coefficients=global_coefficients,
     domain_of_analysis=doa,
)

new_model = simai.models.build(new_conf)
```

## Tests

- smoke tests against Dev
- unit-tests


src/ansys/simai/core/data/global_coefficients_requests.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+class ComputeGlobalCoefficient(GlobalCoefficientRequest):
+    """Computes the result of a Global Coefficient formula."""
+
+    def run(self):
+        """Performs a compute-formula request."""
+        self._client._api.compute_formula(self.project_id, self._calculette_payload)
+
+    def _handle_job_sse_event(self, data):
```
awoimbee, line 173: Would be nice to not redefine `_handle_job_sse_event` many times.

src/ansys/simai/core/data/model_configuration.py
```diff
@@ -0,0 +1,257 @@
...
+                    | *long*: < 1 week
+        continuous: indicates if continuous learning is enabled. Default is False.
+        input: the inputs of the model.
+        output: the outputs of the model.
+        global_coefficients: the Global Coefficients of the model.
+
+    Example:
+        Define a new configuration and launch a build.
+
+        .. code-block:: python
```
awoimbee, line 92: That's a pretty big example, it's gonna get outdated and broken fast...
In fact it's already broken (`simulation_volume` is not defined).

src/ansys/simai/core/data/model_configuration.py
```diff
@@ -0,0 +1,257 @@
...
+
+from dataclasses import asdict, dataclass
+from typing import TYPE_CHECKING, Any, Literal, Optional
+
+if TYPE_CHECKING:
+    from ansys.simai.core.data.projects import Project
+
+
+@dataclass
+class GlobalCoefficientUnit:
```
awoimbee, line 31:
```suggestion
class GlobalCoefficientDefinition:
```
SI `Unit`s could be added later to GCs and that would be confusing

src/ansys/simai/core/data/model_configuration.py
```diff
@@ -0,0 +1,257 @@
...
+
+@dataclass
+class ModelConfiguration:
+    """Configures the build of a model.
+
+    Args:
+        project: the project of the configuration.
+        build_preset: indicates the build duration. Available options:
+
+                    | *debug*: < 30 min
```
awoimbee, line 77:
```suggestion
                    | *debug*: < 30 min, only 4 data
```

src/ansys/simai/core/data/model_configuration.py
```diff
@@ -0,0 +1,257 @@
...
+                output=model_output,
+                global_coefficients=global_coefficients,
+                simulation_volume=simulation_volume,
+            )
+
+            # Launch a mode build with the new configuration
+            new_model = simai.models.build(new_conf)
+    """
+
+    project: "Project" = None
```
awoimbee, line 132:
```suggestion
    project: "Project"
```
Or, but I don't think we want that:
```suggestion
    project: "Optional[Project]" = None
```

src/ansys/simai/core/data/model_configuration.py
```diff
@@ -0,0 +1,257 @@
...
+    output: ModelOutput = ModelOutput()
+
+    def __set_gc(self, gcs: list[tuple | GlobalCoefficientUnit]):
+        verified_gcs = []
+        for gc in gcs:
+            gc_unit = GlobalCoefficientUnit(*gc) if isinstance(gc, tuple) else gc
+            self.project.verify_gc_formula(
+                gc_unit.formula, self.input.boundary_conditions, self.output.surface
+            )
+            verified_gcs += [gc_unit]
```
awoimbee, line 150:
```suggestion
            verified_gcs.append(gc_unit)
```

src/ansys/simai/core/data/model_configuration.py
```diff
@@ -0,0 +1,257 @@
...
+        self.__dict__["global_coefficients"] = verified_gcs
+
+    def __get_gc(self):
+        return self.__dict__.get("global_coefficients")
+
+    global_coefficients: ModelOutput = property(__get_gc, __set_gc)
```
awoimbee, line 156:
```suggestion
    global_coefficients: Optional[list[GlobalCoefficientUnit]] = property(__get_gc, __set_gc)
```
^ remove optional if you default to an empty list instead of None
Btw it's not pretty but seems like we're limited by what dataclass can do here

src/ansys/simai/core/data/model_configuration.py
```diff
@@ -0,0 +1,257 @@
...
+        for gc in gcs:
+            gc_unit = GlobalCoefficientUnit(*gc) if isinstance(gc, tuple) else gc
+            self.project.verify_gc_formula(
+                gc_unit.formula, self.input.boundary_conditions, self.output.surface
+            )
+            verified_gcs += [gc_unit]
+        self.__dict__["global_coefficients"] = verified_gcs
+
+    def __get_gc(self):
+        return self.__dict__.get("global_coefficients")
```
awoimbee, line 154: Imo we can make this a bit nicer for free:
```suggestion
        return self.__dict__.setdefault("global_coefficients", [])
```

src/ansys/simai/core/data/model_configuration.py
```diff
@@ -0,0 +1,257 @@
...
+
+    global_coefficients: ModelOutput = property(__get_gc, __set_gc)
+
+    def __init__(
+        self,
+        boundary_conditions: Optional[dict[str, Any]] = None,
+        build_preset: Optional[str] = None,
+        continuous: Optional[bool] = False,
+        fields: Optional[dict[str, Any]] = None,
+        global_coefficients: Optional[dict[dict[str, Any]] | GlobalCoefficientUnit] = None,
```
awoimbee, line 164: Wait, you have a custom set for GCs that accepts `tuple` but here you have another custom logic based on `dict` ?
What are users supposed to use ?
BTW:
```suggestion
        global_coefficients: Optional[list[dict[dict[str, Any]] | GlobalCoefficientUnit]] = None,
```
Ans same as above, we could default to an empty list

src/ansys/simai/core/data/model_configuration.py
```diff
@@ -0,0 +1,257 @@
...
+        simulation_volume: Optional[dict[str, Any]] = None,
+        project: Optional["Project"] = None,
+        input: Optional[ModelInput] = None,
+        output: Optional[ModelOutput] = None,
+    ):
+        """Sets the properties of a build configuration."""
+        self.project = project
```
awoimbee, line 171: so self.project is None and stuff like `self.project.compute_gc_formula` crashes

src/ansys/simai/core/data/model_configuration.py
```diff
@@ -0,0 +1,257 @@
...
+        if boundary_conditions is not None:
+            self.input.boundary_conditions = list(boundary_conditions.keys())
```
awoimbee, line 173: We ask for a dict and they only use the keys, shouild we ask for a list ?

src/ansys/simai/core/data/model_configuration.py
```diff
@@ -0,0 +1,257 @@
...
+        self.build_preset = build_preset
+        self.continuous = continuous
+        if fields is not None:
```
awoimbee, line 176: fields is not documented

src/ansys/simai/core/data/model_configuration.py
```diff
@@ -0,0 +1,257 @@
...
+            "fields": flds,
+            "global_coefficients": gcs,
+            "simulation_volume": self.simulation_volume,
+        }
+
+    def compute_global_coefficient(self):
+        """Computes the results of the formula for all Global Coefficients with respect to the project's sample."""
+
+        return [
+            self.project.compute_gc_formula(
```
awoimbee, line 253: This should be parallelized

awoimbee (commented):
Cool but it requires more work

src/ansys/simai/core/data/global_coefficients_requests.py
```diff
@@ -0,0 +1,209 @@
...
+if TYPE_CHECKING:
+    import ansys.simai.core.client
+
+EXTRA_CALCULETTE_FIELDS = ["Area", "Normals", "Centroids"]
+
+logger = logging.getLogger(__name__)
+
+
+class GlobalCoefficientRequest(ABC, ComputableDataModel):
+    """Creates the basic request for making requests related to Global Coefficients."""
```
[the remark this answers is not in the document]
romane-m: """Creates the foundational request for subsequent Global Coefficients requests/inquiries."""

src/ansys/simai/core/data/global_coefficients_requests.py
```diff
@@ -0,0 +1,209 @@
...
+            ] = f"Verification of formula {target.get('formula')} failed with {data.get('reason')}"
+            logger.debug(
+                f"Verification of formula {target.get('formula')} failed with {data.get('reason')}"
+            )
+            self._set_has_failed()
+            logger.error(self.fields.get("error"))
+
+
+class CheckGlobalCoefficientDirectory(GlobalCoefficientRequestDirectory[CheckGlobalCoefficient]):
+    """Extends GlobalCoefficientRequestDirectory for verifying a Global Coefficient formuala."""
```
romane-m, line 161: formula

src/ansys/simai/core/data/model_configuration.py
```diff
@@ -0,0 +1,257 @@
...
+from dataclasses import asdict, dataclass
+from typing import TYPE_CHECKING, Any, Literal, Optional
+
+if TYPE_CHECKING:
+    from ansys.simai.core.data.projects import Project
+
+
+@dataclass
+class GlobalCoefficientUnit:
+    """Single Global Coefficient field.
```
romane-m, line 32: Global coefficient definition/parameter field.

romane-m (commented):
1st part of my review 
Some edits still in progress

src/ansys/simai/core/data/model_configuration.py
yias: IMO, `fields` should not be documented, as they are not supposed to be used by clients/customers.

src/ansys/simai/core/data/model_configuration.py
yias: I don't see an immediate reason to have them parallelized. It seems to me that parallelization at this point would be "good to have" than "should have". If this results in a bottleneck in future, we could improve it then.

src/ansys/simai/core/data/projects.py
```diff
@@ -138,6 +147,66 @@ def get_variables(self) -> dict[str, list[str]] | None:
             data[key] = [local_field.get("name") for local_field in local_fields]
         return data
 
+    def verify_gc_formula(
+        self, gc_formula: str, bc: list[str] = None, surface_variables: list[str] = None
+    ):
+        """Verifies if the Global Coefficient formula is valid."""
+
+        if not self.sample:
+            raise ProcessingError(
```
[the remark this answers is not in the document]
yias: an error is thrown

src/ansys/simai/core/api/project.py
```diff
@@ -91,3 +91,39 @@ def launch_build(
...
+                                formula that describes a Global Coefficient
+        """
+
+        return self._post(
+            f"projects/{project_id}/compute-formula",
+            json=calculette_payload,
+            return_json=False,
+        )
+
+    def request_calculette_session(self, project_id: str):
```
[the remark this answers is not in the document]
yias: `calcullete` is a backend service. Some times a dedicated session should be requested for a project. This function does that. Nevertheless, I removed it as it is not needed for this feature.

src/ansys/simai/core/data/model_configuration.py
yias: good point, I changed to `GlobalCoefficientDefinition` to better guide users.

src/ansys/simai/core/data/model_configuration.py
yias: the idea here is to have it backwards compatible with the `last_model_configuration`, where the `ModelConfiguration` should consume what it gets from the server too. This is why it checks if `boundary_conditions` is not `None`, to check it is the user's input or the server response.

src/ansys/simai/core/data/model_configuration.py
yias: Correct, which it is fair considering that a build configuration is tight to a project. Regardless, I set some checks and raise exception in `compute_gc_formula` for highlighting it to the user.

src/ansys/simai/core/data/global_coefficients_requests.py
yias: it is unavoidable in this point as the servers responses are different between the `check_formula` and the `compute_formula` events.

src/ansys/simai/core/data/model_configuration.py
yias: I merged with `doa` so that the example is up-to-date. True that is big, however, moving into examples could be done in a later stage.

src/ansys/simai/core/data/model_configuration.py
yias: `get` looks more descriptive to me.

src/ansys/simai/core/data/global_coefficients_requests.py
```diff
@@ -0,0 +1,209 @@
...
+    def run(self):
+        """Abstract method to perform the customized required process."""
+
+
+GlobalCoefficientRequestType = TypeVar(
+    "GlobalCoefficientRequestType", bound=GlobalCoefficientRequest
+)
+
+
+class GlobalCoefficientRequestDirectory(Directory[GlobalCoefficientRequestType]):
```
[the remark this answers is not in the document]
yias: `CheckGlobalCoefficientDirectory` and `ComputeGlobalCoefficientDirectory` follow the same pattern and nomenclature of the rest of the code in interfaces that handle server requests. This is to secure coherency and stability. They are directories of SSE events too. I see your point regarding the naming (if I understand correctly), but any refactoring of this type could come to a later stage.

src/ansys/simai/core/data/model_configuration.py
awoimbee: Ok but why is project optional then ? (`project: Optional["Project"] = None`)

yias: to give more flexibility to the user to gradually compose a configuration, e.g., first create inputs, then outputs, etc. To check the gcs, however, a project should be defined.

awoimbee:
I will let someone else than me review this PR since I already put so many comments on it

src/ansys/simai/core/data/global_coefficients_requests.py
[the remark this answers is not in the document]
tmpbeing: > CheckGlobalCoefficientDirectory and ComputeGlobalCoefficientDirectory follow the same pattern and nomenclature of the rest of the code in interfaces that handle server requests. 

We talked about it a bit and I agree that it's fine as is for now but I wouldn't agree with the above statement.
Directory is a class built for server objects with public CRUD operations. Currently our SSE behaviors are tied to it because we never had a need for SSE outside of it. 
But there is in the code base at least an example of a class similar to directory but without the data model: https://github.com/ansys/pysimai/blob/df387bfcab19ad45efbd911384aa0e882eca01ec/src/ansys/simai/core/data/design_of_experiments.py#L34
For a refactoring we could move the SSE behaviors to a separate class and then compose where needed: on what could become the `GlobalCoefficientCollection` (or something else) and on the existing directories. Maybe we'll also need to do the same for the SSE part of `ComputableDataModel`.

src/ansys/simai/core/data/global_coefficients_requests.py
```diff
@@ -0,0 +1,209 @@
...
+        if state in PENDING_STATES:
+            logger.debug(f"{self._classname} id {self.id} set status pending")
+            self._set_is_pending()
+        elif state == "successful":
+            logger.debug(f"{self._classname} id {self.id} set status successful")
+            self._set_is_over()
+        elif state in ERROR_STATES:
+            self.fields[
+                "error"
+            ] = f"Verification of formula {target.get('formula')} failed with {data.get('reason')}"
```
tmpbeing, line 152: You're using the message 3 times (twice copied then you get it to log the error), you could put it in a variable instead.
Also wondering if we should put a default in the `data.get("reason")` but I don't have a good idea of default error message.

src/ansys/simai/core/data/global_coefficients_requests.py
```diff
@@ -0,0 +1,209 @@
...
+            logger.debug(f"{self._classname} id {self.id} set status pending")
+            self._set_is_pending()
+        elif state == "successful":
+            logger.debug(f"{self._classname} id {self.id} set status successful")
+            self._result = data.get("result").get("value")
+            self._set_is_over()
+        elif state in ERROR_STATES:
+            self.fields[
+                "error"
+            ] = f"Verification of formula {target.get('formula')} failed with {data.get('reason')}"
```
tmpbeing, line 191: Same comment

tmpbeing, line 191:
```suggestion
            ] = f"Computation of global coefficient {target.get('formula')} failed with {data.get('reason')}"
```

src/ansys/simai/core/data/model_configuration.py
```diff
@@ -0,0 +1,434 @@
...
+
+            # Define the inputs of the model
+            model_input = ModelInput(surface=["Velocity"], boundary_conditions=["Vx"])
+
+            # Define the outputs of the model
+            model_output = ModelOutput(
+                surface=["Pressure", "WallShearStress_0"], volume=["Velocity_0", "Pressure"]
+            )
+
+            # Define the model coefficient
```
tmpbeing, line 240:
```suggestion
            # Define the model coefficients
```

src/ansys/simai/core/data/model_configuration.py
```diff
@@ -0,0 +1,434 @@
...
+    input: ModelInput = field(default_factory=lambda: ModelInput())
+    output: ModelOutput = field(default_factory=lambda: ModelOutput())
+    domain_of_analysis: DomainOfAnalysis = field(default_factory=lambda: DomainOfAnalysis())
+
+    def __set_gc(self, gcs: list[GlobalCoefficientDefinition]):
+        verified_gcs = []
+
+        for gc in gcs:
+            gc_unit = GlobalCoefficientDefinition(*gc) if isinstance(gc, tuple) else gc
+            if self.project is not None:
```
tmpbeing, line 282: So if I don't set project and set GC after creating my config, my formulas won't be verified, correct ?
I think this is potential issue as right now the backend doesn't check the formulas, they're assumed to be correct/verified by the client (which might not be a great assumption but it's convenient for now).

src/ansys/simai/core/data/projects.py
```diff
@@ -126,7 +135,7 @@ def is_trainable(self) -> bool:
         tt = self._client._api.is_project_trainable(self.id)
         return IsTrainableInfo(**tt)
 
-    def get_variables(self) -> Optional[Dict[str, List[str]]]:
+    def get_variables(self) -> Optional[dict[str, List[str]]]:
```
tmpbeing, line 138: I think you might have to rebase for the tests to work properly, if I remember correctly this isn't valid in Python 3.9 which we're supposed to support (the typing part).

tmpbeing (commented):
LGTM mostly, good work.
Just worried about 3.9 compatibility and a possible way to send broken models if we don't check formulas.

src/ansys/simai/core/data/model_configuration.py
yias: would it be preferable if 'project' was required variable?

yias: so I made `project` a required argument for model configuration. I also introduced an exception there, to enforce that a project should be defined.

src/ansys/simai/core/data/projects.py
yias: I think I corrected it.

src/ansys/simai/core/data/global_coefficients_requests.py
yias: I set a `error_message` variable and put `UNKNOWN ERROR` as the default for `reason`.

src/ansys/simai/core/data/global_coefficients_requests.py
yias: corrected. resolving comment

tmpbeing (approved):
:+1:

PR #40, opened 2024-04-15
fix: use the private optimization _try_geometry

tmpbeing:
Method was made private when moving to PySimAI but the call was never updated.


yias (approved):
looks good.

PR #42, opened 2024-04-15
Allow user to set Train/Validation/Test subset on training data

kliment-slice:
[sc-17795]
set_subset method added to TrainingData class, subset enum type. Tested and verified.

### DIFF
--- a/src/ansys/simai/core/api/training_data.py
+++ b/src/ansys/simai/core/api/training_data.py
@@ -24,6 +24,7 @@ from typing import Any, Dict, Iterable, Optional
 
 from ansys.simai.core.api.mixin import ApiClientMixin
 from ansys.simai.core.data.types import SubsetEnum
+from ansys.simai.core.errors import InvalidArguments
 
 
 class TrainingDataClientMixin(ApiClientMixin):
@@ -71,8 +72,7 @@ class TrainingDataClientMixin(ApiClientMixin):
         self, project_id: str, training_data_id: str, subset: SubsetEnum
     ) -> None:
         if subset not in SubsetEnum.__members__.values():
-            raise ValueError("Must be one of: Ignored, Training, Test, Validation.")
-        else:
-            return self._put(
-                f"projects/{project_id}/data/{training_data_id}/subset", json={"subset": subset}
-            )
+            raise InvalidArguments("Must be one of: Ignored, Training, Test, Validation.")
+        return self._put(
+            f"projects/{project_id}/data/{training_data_id}/subset", json={"subset": subset}
+        )
--- a/src/ansys/simai/core/api/training_data.py
+++ b/src/ansys/simai/core/api/training_data.py
@@ -23,8 +23,6 @@
 from typing import Any, Dict, Iterable, Optional
 
 from ansys.simai.core.api.mixin import ApiClientMixin
-from ansys.simai.core.data.types import SubsetEnum
-from ansys.simai.core.errors import InvalidArguments
 
 
 class TrainingDataClientMixin(ApiClientMixin):
@@ -68,11 +66,7 @@ class TrainingDataClientMixin(ApiClientMixin):
     def get_training_data_subset(self, project_id: str, training_data_id: str) -> Dict[str, Any]:
         return self._get(f"projects/{project_id}/data/{training_data_id}/subset")
 
-    def put_training_data_subset(
-        self, project_id: str, training_data_id: str, subset: SubsetEnum
-    ) -> None:
-        if subset not in SubsetEnum.__members__.values():
-            raise InvalidArguments("Must be one of: Ignored, Training, Test, Validation.")
+    def put_training_data_subset(self, project_id: str, training_data_id: str, subset: str) -> None:
         return self._put(
             f"projects/{project_id}/data/{training_data_id}/subset", json={"subset": subset}
         )
--- a/src/ansys/simai/core/api/training_data.py
+++ b/src/ansys/simai/core/api/training_data.py
@@ -23,6 +23,7 @@
 from typing import Any, Dict, Iterable, Optional
 
 from ansys.simai.core.api.mixin import ApiClientMixin
+from ansys.simai.core.data.types import SubsetEnum
 
 
 class TrainingDataClientMixin(ApiClientMixin):
@@ -66,7 +67,12 @@ class TrainingDataClientMixin(ApiClientMixin):
     def get_training_data_subset(self, project_id: str, training_data_id: str) -> Dict[str, Any]:
         return self._get(f"projects/{project_id}/data/{training_data_id}/subset")
 
-    def put_training_data_subset(self, project_id: str, training_data_id: str, subset: str) -> None:
-        return self._put(
-            f"projects/{project_id}/data/{training_data_id}/subset", json={"subset": subset}
-        )
+    def put_training_data_subset(
+        self, project_id: str, training_data_id: str, subset: SubsetEnum
+    ) -> None:
+        if subset not in SubsetEnum.__members__.values():
+            raise ValueError("Must be one of: Ignored, Training, Test, Validation.")
+        else:
+            return self._put(
+                f"projects/{project_id}/data/{training_data_id}/subset", json={"subset": subset}
+            )
--- a/src/ansys/simai/core/api/training_data.py
+++ b/src/ansys/simai/core/api/training_data.py
@@ -65,3 +65,8 @@ class TrainingDataClientMixin(ApiClientMixin):
 
     def get_training_data_subset(self, project_id: str, training_data_id: str) -> Dict[str, Any]:
         return self._get(f"projects/{project_id}/data/{training_data_id}/subset")
+
+    def put_training_data_subset(self, project_id: str, training_data_id: str, subset: str) -> None:
+        return self._put(
+            f"projects/{project_id}/data/{training_data_id}/subset", json={"subset": subset}
+        )
--- a/src/ansys/simai/core/data/training_data.py
+++ b/src/ansys/simai/core/data/training_data.py
@@ -76,7 +76,7 @@ class TrainingData(ComputableDataModel):
             for training_data_part in self.fields["parts"]
         ]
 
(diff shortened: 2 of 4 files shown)


src/ansys/simai/core/api/training_data.py
```diff
@@ -65,3 +65,8 @@ def compute_training_data(self, training_data_id: str) -> None:
 
     def get_training_data_subset(self, project_id: str, training_data_id: str) -> Dict[str, Any]:
         return self._get(f"projects/{project_id}/data/{training_data_id}/subset")
+
+    def put_training_data_subset(self, project_id: str, training_data_id: str, subset: str) -> None:
```
yias, line 69: nit: though it's not a practice that is followed in this module, a docstring or/and args explanation would be good.

src/ansys/simai/core/data/training_data.py
```diff
@@ -88,6 +89,11 @@ def get_subset(self, project: Identifiable["Project"]) -> Optional[str]:
         project_id = get_id_from_identifiable(project, default=self._client._current_project)
         return self._client._api.get_training_data_subset(project_id, self.id).get("subset")
 
+    def set_subset(self, project: Identifiable["Project"], subset: SubsetEnum) -> None:
```
yias, line 92: 1. nit: you could consider `make_subset` as a name of the method too.

2. for this function an `Args` section is needed, similar to `get_subset` that it would be consumed by the documentation. @romane-m may have suggestions on this too.

src/ansys/simai/core/data/types.py
```diff
@@ -262,3 +263,10 @@ def get_object_from_identifiable(
         return get_object_from_identifiable(default, directory)
     else:
         raise InvalidArguments(f"Argument {identifiable} is neither a data model nor an ID string.")
+
+
+class SubsetEnum(str, Enum):
```
yias, line 268: 1. Not sure I understand the point of `none` 😕 Does it mean nothing in the body of the request? 
2. It's generally used to have upper case in enumerations:

```py
TRAINING = "Training"
VALIDATION = "Validation"
TEST = "TEST"
```
3. Is it expected that users would select one of those for creating a subset? In this case, should this enumeration be exposed to the users, so that they pull it, something like 
```py
sbts = simai.training_data.subset_names
td.set_subset("wc2y34", sbts.VALIDATION)
```

In this case, class `TrainingDataDirectory` should have a property for it

yias (commented):
just a few comments from my side :)

src/ansys/simai/core/data/training_data.py
romane-m, line 92: The method name seems confusing to me, maybe we should be more specific. If we are defining a type for an existing subset of data, I would suggest to rename the method to integrate the word **type** in it.

src/ansys/simai/core/data/training_data.py
```diff
@@ -88,6 +89,11 @@ def get_subset(self, project: Identifiable["Project"]) -> Optional[str]:
...
+        """Declare a subset for training data in relation to given project."""
```
romane-m, line 93: Define how a subset of training data will be interpreted by the model.

src/ansys/simai/core/data/training_data.py
awoimbee: This is confusing because we're in the `TrainingData` object but we're editing the `ModelTrainingDataAssociation`.
Maybe it could be named something like `set_project_training_subset` ?

tests/test_training_data.py
```diff
@@ -103,3 +105,17 @@ def test_get_subset(training_data_factory, project_factory, td_factory_args):
...
+    project = project_factory(id="n07e45y", name="bananarama")
+    td: TrainingData = training_data_factory(project=project, subset=SubsetEnum.training)
+
+    responses.add(
+        responses.GET,
+        f"https://test.test/projects/{project.id}/data/{td.id}/subset",
+        status=200,
+        json={"subset": SubsetEnum.validation},
+    )
+    assert td.get_subset(project=project) == SubsetEnum.validation
```
awoimbee, line 121: it's called `test_set_subset`...

src/ansys/simai/core/data/training_data.py
```diff
@@ -88,6 +89,20 @@ def get_subset(self, project: Identifiable["Project"]) -> Optional[str]:
         project_id = get_id_from_identifiable(project, default=self._client._current_project)
         return self._client._api.get_training_data_subset(project_id, self.id).get("subset")
 
+    def assign_subset(self, project: Identifiable["Project"], subset: SubsetEnum) -> None:
+        """Assign the data subset in relation to a given project.
+
+        Args:
+            project: ID or :class:`model <.projects.Project>` of the project to check
+                the :class:`~.projects.Project` object for, or its ID.
```
awoimbee, line 97: :eyes:

src/ansys/simai/core/data/training_data.py
```diff
@@ -269,3 +284,12 @@ def upload_folder(
         if compute:
             self._client._api.compute_training_data(training_data_id)
         return uploaded_parts
+
+    @property
+    def subset_names(self) -> List[str]:
```
awoimbee, line 289: That's not needed since we've got the `SubsetEnum` type.
I think you wrote this because of a previous comment that was misguided by the fact that when reading the previous code, reviewers (me included) had the impressions that subsets where being created and not just assigned (hense the need to list them) -> it's not the case so we don't need this function.

tests/test_training_data.py
```diff
@@ -103,3 +105,17 @@ def test_get_subset(training_data_factory, project_factory, td_factory_args):
...
+    project = project_factory(id="n07e45y", name="bananarama")
+    td: TrainingData = training_data_factory(project=project, subset=SubsetEnum.TRAINING)
+
+    responses.add(
+        responses.GET,
+        f"https://test.test/projects/{project.id}/data/{td.id}/subset",
+        status=200,
+        json={"subset": SubsetEnum.VALIDATION},
+    )
+    assert td.get_subset(project=project) == SubsetEnum.VALIDATION
```
awoimbee, line 121: Feels like I already wrote that comment before:
we're in `test_set_subset`, why are we testing `td.get_subset` ?

src/ansys/simai/core/data/types.py
awoimbee, line 268: BTW shouldn't `get_subset` return a `SubsetEnum` instead of an `str` ?

src/ansys/simai/core/data/training_data.py
yias: If the intention is to use it like:

```py
from ansys.simai.core.data.types import SubsetEnum

td.assign_subset("project_id", SubsetEnum.VALIDATION)
```

then, indeed, we don't need to export it through a `TrainingData` property. My comment was more on securing the intended usage, than requesting a code modification 😅

src/ansys/simai/core/api/training_data.py
```diff
@@ -65,3 +66,13 @@ def compute_training_data(self, training_data_id: str) -> None:
 
     def get_training_data_subset(self, project_id: str, training_data_id: str) -> Dict[str, Any]:
         return self._get(f"projects/{project_id}/data/{training_data_id}/subset")
+
+    def put_training_data_subset(
+        self, project_id: str, training_data_id: str, subset: SubsetEnum
+    ) -> None:
+        if subset not in SubsetEnum.__members__.values():
+            raise ValueError("Must be one of: Ignored, Training, Test, Validation.")
```
yias, line 74: It seems this should be a `ansys.simai.core.errors.InvalidArguments` exception.

src/ansys/simai/core/api/training_data.py
```diff
@@ -65,3 +66,13 @@ def compute_training_data(self, training_data_id: str) -> None:
...
+        else:
```
yias, line 75: nit: no need `else` here

yias (approved):
just one more thing for the exception, but it looks fine to me.

src/ansys/simai/core/api/training_data.py
```diff
@@ -65,3 +67,12 @@ def compute_training_data(self, training_data_id: str) -> None:
 
     def get_training_data_subset(self, project_id: str, training_data_id: str) -> Dict[str, Any]:
         return self._get(f"projects/{project_id}/data/{training_data_id}/subset")
+
+    def put_training_data_subset(
+        self, project_id: str, training_data_id: str, subset: SubsetEnum
+    ) -> None:
+        if subset not in SubsetEnum.__members__.values():
+            raise InvalidArguments("Must be one of: Ignored, Training, Test, Validation.")
```
awoimbee, line 75: IMO this is the raw API interface, so in here take in `subset: str` and don't check `if subset not in SubsetEnum.__members__.values():`, this should be done in `def assign_subset`

src/ansys/simai/core/data/training_data.py
```diff
@@ -75,18 +76,34 @@ def parts(self) -> List["TrainingDataPart"]:
...
 
         Returns:
-            Name of the subset that the training data belongs to in the given project.
+            SubsetEnum of the subset that the training data belongs to in the given project.
+                (e.g. <SubsetEnum.VALIDATION: 'Validation'>)
         """
         project_id = get_id_from_identifiable(project, default=self._client._current_project)
-        return self._client._api.get_training_data_subset(project_id, self.id).get("subset")
+        subset_value = self._client._api.get_training_data_subset(project_id, self.id).get("subset")
+        return SubsetEnum(subset_value)
```
awoimbee, line 92: Will crash if `subset_value` is `None`, which it is for all new data until a training is launched.

src/ansys/simai/core/data/types.py
awoimbee, line 268: If it works, disregard, if you're having issues, know that it's because [the behavior changed since python3.11](https://github.com/python/cpython/issues/100458) and that the [workaround](https://github.com/equinor/webviz-ert/pull/449/files) looks like that.

Since it's starting to be complicated, keep in mind we can just create a type `SubSetEnum = Literal["Ignored", "Training", "Validation", "Test"]` instead

awoimbee (approved):
Perfect !

PR #44, opened 2024-04-18
Set DoA in model configuration

yias:
## Description

This PR enables the definition of Domain of Analysis from scratch. 

story: [sc-18078]

NB: it also updates the doc styles to vale v3, see [here](https://actions.docs.ansys.com/version/stable/migrations/docs-style-vale-version-update.html).

## Implementation

- The class `DomainAxisDefinition` represents a single field of DoA (e.g., the length, the width or the height). The param `position` corresponds to the key `type` for being aligning with the terminology in the UI and avoid confusion with the term `type`. A validation of the param `value` is implemented for securing the proper value is filled in according to `position`.
- The class `DomainOfAnalysis` contains the components of DoA, i.e., the length, the width or the height, as type of `DomainAxisDefinition`
- The class `ModelConfiguration` has an attribute `domain_of_analysis`; 
  - it constructor can accept a json configuration as retrieved from the server, to secure backwards compatibility. In this case, the attribute `domain_of_analysis` is being constructed from the `simulation_volume` property. The constructor can accept either a `domain_of_analysis` or a 'simulation_volume' parameter; in the case where both are defined it raises an exception.
  - besides from the construction, the `domain_of_analysis` can be assigned directly as a property. 
  - a protected method `_to_payload` is introduced to serialize a `ModelConfiguration` object to the payload of a launch-build request

## Usage

```py
from ansys.simai.core.data.models import DomainOfAnalysis

# Get the last configuration from a project
bld_conf = my_project.last_model_configuration

# Create a new DoA
new_doa = DomainOfAnalysis(
    height=("relative_to_min", 4.5, 4.1),
    length=("relative_to_center", 1.5, 150),
    width=("absolute", -1.5, 6.1),
)

# Assign the new DoA to the configuration

bld_conf.domain_of_analysis = new_doa
```

## Tests

- smoke tests against Dev
- unit-tests


src/ansys/simai/core/data/models.py
```diff
@@ -38,6 +114,43 @@ class ModelConfiguration:
     simulation_volume: Dict[str, Any] = None
     project_id: str = None
 
+    @property
+    def domain_of_analysis(self) -> DomainOfAnalysis | None:
+        """The Domain of Analysis of the model configuration."""
+        if not self.simulation_volume:
+            return None
```
yias, line 121: I'm wondering if returning an empty `DomainOfAnalysis` object would be better than just `None`.

src/ansys/simai/core/data/models.py
```diff
@@ -21,9 +21,85 @@
...
+
+            # get DoA from the configuration
+            doa = bld_conf.domain_of_analysis
+
+            # create a new DoA
+            new_doa = simai.models.domain_of_analysis()
+
+            new_doa.Height = simai.models.doa_axis_definition("relative_to_min", 4.5, 4.1)
+            new_doa.Length = simai.models.doa_axis_definition("relative_to_center", 0.5, 15)
+            new_doa.Width = simai.models.doa_axis_definition("absolute", -1.5, 6.1)
```
tmpbeing, line 92: Creating an empty object and populating it seems fairly unconventional in python.
Why not just use the `__init__` instead ?
You could even have it so you don't need the simai.models.doa_axis_definition by creating them in the `__init__`
You'd end up with something like:
`new_doa = DomainOfAnalysis(height=("relative_to_min, 4.5, 4.1), length=("relative_to_center", 0.5, 15), width=("absolute", -1.5, 6.1))`
Way less verbose for the user and he doesn't need to know about `doa_axis_definition` or `DomainAxisDefinition` itself.

src/ansys/simai/core/data/models.py
```diff
@@ -38,6 +114,43 @@ class ModelConfiguration:
     simulation_volume: Dict[str, Any] = None
     project_id: str = None
 
+    @property
+    def domain_of_analysis(self) -> DomainOfAnalysis | None:
```
tmpbeing, line 118:
```suggestion
    simulation_volume: Dict[str, Any] = None
    project_id: str = None

    @property
    def domain_of_analysis(self) -> Optional[DomainOfAnalysis]
```
The `|` operator was added in 3.10 and we support from 3.9. The worrying part is that the CI passed.

src/ansys/simai/core/data/models.py
```diff
@@ -38,6 +114,43 @@ class ModelConfiguration:
...
+
+        self.simulation_volume["X"] = self._set_doa_axis(doa.Length, "Length")
+        self.simulation_volume["Y"] = self._set_doa_axis(doa.Width, "Width")
+        self.simulation_volume["Z"] = self._set_doa_axis(doa.Height, "Height")
+
+    def _get_doa_axis(self, rel_pos: str) -> DomainAxisDefinition:
+        pos = self.simulation_volume.get(rel_pos)
+        return DomainAxisDefinition(position=pos["type"], value=pos["value"], length=pos["length"])
+
+    def _set_doa_axis(self, fld: DomainAxisDefinition, param: str) -> Dict[str, Any]:
```
awoimbee, line 146: Why not directly use the serialization/deserialization features of pydantic ?

if `fld` can be `None` then its type should be `Optional[fld]` (is `fld` shorthand for "field" ?)

src/ansys/simai/core/data/models.py
```diff
@@ -120,3 +233,24 @@ def build(
                 dismiss_data_with_volume_overflow,
             )
         )
+
+    def doa_axis_definition(
```
awoimbee, line 237: @tmpbeing I saw your comment above:
> I think not raising a ValidationError would be nice (use custom errors wherever we can)

But IMO writing functions just to try/except like this is bloat, maybe we can make `InvalidArguments` subclass `ValidationError` ? Or just reexport it as `ansys.simai.core.ValidationError` ?

tests/test_models.py
```diff
@@ -138,3 +142,51 @@ def test_build(simai_client):
...
+
+    model_conf = ModelConfiguration(project_id=MODEL_RAW["project_id"], **MODEL_CONF_RAW)
+
+    doa = model_conf.domain_of_analysis
+
+    new_height = {"position": "relative_to_center", "value": 0.5, "length": 15.2}
+
+    doa.Height = simai_client.models.doa_axis_definition(**new_height)
+
+    model_conf.domain_of_analysis = doa
```
awoimbee, line 160: When I set `model_conf.domain_of_analysis.height = DomainAxisDefinition(...)` I expect `model_conf` to be updated !
This to me really needs to be solved.

awoimbee (changes requested):

src/ansys/simai/core/data/models.py
```diff
@@ -21,9 +21,85 @@
...
+
+    position: Literal["relative_to_min", "relative_to_max", "relative_to_center", "absolute"]
+    value: float
+    length: PositiveFloat
+
+    @model_validator(mode="after")
+    def chek_value(self):
+        """Evaluates that the value is set correctly according to the position argument."""
+        if self.value < 0 and self.position != "absolute":
+            raise InvalidArguments(
```
romane-m, line 61: "value must be a positive number when the position is not absolute."

maybe add some formatting to the arguments' names?

src/ansys/simai/core/data/models.py
```diff
@@ -21,9 +21,85 @@
...
+                "value should be a positive number when the position is not absolute"
+            )
+        return self
+
+
+@dataclass
+class DomainOfAnalysis:
+    """Definition of the Domain of Analysis.
```
romane-m, line 69: Defines the Domain of Analysis.

src/ansys/simai/core/data/models.py
```diff
@@ -21,9 +21,85 @@
...
+
+    Args:
```
romane-m, line 71: arguments in lower case: 

 Args:
        length: Domain of Analysis definition along the X axis.
        width: Domain of Analysis definition along the Y axis.
        height: Domain of Analysis definition along the Z axis.

src/ansys/simai/core/data/models.py
```diff
@@ -21,9 +21,85 @@
...
+        Length: the definition of the Domain of Analysis along the X axis
+        Width: the definition of the Domain of Analysis along the Y axis
+        Height: the definition of the Domain of Analysis along the Z axis
+
+    Example:
+        Get the Domain of Analysis from a configuration and replace it with a new one
+
+        .. code-block:: python
+
+            # get the last configuration from a project
```
romane-m, line 81: Get

src/ansys/simai/core/data/models.py
```diff
@@ -21,9 +21,85 @@
...
+            bld_conf = my_project.last_model_configuration
+
+            # get DoA from the configuration
```
romane-m, line 84: Get

src/ansys/simai/core/data/models.py
```diff
@@ -21,9 +21,85 @@
...
+            doa = bld_conf.domain_of_analysis
+
+            # create a new DoA
```
romane-m, line 87: Create

src/ansys/simai/core/data/models.py
```diff
@@ -21,9 +21,85 @@
...
+            new_doa = simai.models.domain_of_analysis()
+
+            new_doa.Height = simai.models.doa_axis_definition("relative_to_min", 4.5, 4.1)
+            new_doa.Length = simai.models.doa_axis_definition("relative_to_center", 0.5, 15)
+            new_doa.Width = simai.models.doa_axis_definition("absolute", -1.5, 6.1)
+
+            # set the new DoA to the configuration
```
romane-m, line 94: Assign the new DoA to the configuration

src/ansys/simai/core/data/models.py
```diff
@@ -21,9 +21,85 @@
 # SOFTWARE.
 
 from dataclasses import asdict, dataclass
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Literal
```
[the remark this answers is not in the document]
yias: yup, good point.

src/ansys/simai/core/data/models.py
```diff
@@ -38,6 +114,43 @@ class ModelConfiguration:
...
+
+        doa = DomainOfAnalysis()
+
+        doa.Length = self._get_doa_axis("X")
+        doa.Width = self._get_doa_axis("Y")
+        doa.Height = self._get_doa_axis("Z")
+
+        return doa
+
+    @domain_of_analysis.setter
```
romane-m, line 131: A docstring is not required for a setter following a property as the setter only exposes set and get methods.
Examples can be included though.

src/ansys/simai/core/data/models.py
```diff
@@ -38,6 +114,43 @@ class ModelConfiguration:
...
+
+    def _get_doa_axis(self, rel_pos: str) -> DomainAxisDefinition:
+        pos = self.simulation_volume.get(rel_pos)
+        return DomainAxisDefinition(position=pos["type"], value=pos["value"], length=pos["length"])
+
+    def _set_doa_axis(self, fld: DomainAxisDefinition, param: str) -> Dict[str, Any]:
+        """Serializes a DomainAxisDefinition to the required format for the server."""
+        if fld is None:
+            raise InvalidArguments(
+                f"Empty parameter {param} found when setting the Domain of Analysis. All the parameters should be filled in."
```
romane-m, line 150: f"Empty parameter {param} found when setting the Domain of Analysis. All parameters should be set."

src/ansys/simai/core/data/models.py
```diff
@@ -120,3 +233,24 @@ def build(
                 dismiss_data_with_volume_overflow,
             )
         )
+
+    def doa_axis_definition(
+        self, position: str, value: float, length: PositiveFloat
+    ) -> DomainAxisDefinition:
+        """Sets an axis in the Domain of Analysis.
```
romane-m, line 240: """Set an axis in the Domain of Analysis.

src/ansys/simai/core/data/models.py
```diff
@@ -120,3 +233,24 @@ def build(
...
+
+        Returns:
+            a :class:`DomainAxisDefinition <DomainAxisDefinition>` object.
+
+        Raises:
+            `ansys.simai.core.errors.InvalidArguments`: when the arguments are not compatible with the definition of an axis of the Domain of Analysis.
```
romane-m, line 246: `ansys.simai.core.errors.InvalidArguments`: When the arguments are not compatible with the definition of an axis within the Domain of Analysis.

tests/test_models.py
```diff
@@ -138,3 +142,51 @@ def test_build(simai_client):
     model_conf.pop("project_id")
 
     assert model_conf == MODEL_CONF_RAW
+
+
+def test_set_doa(simai_client):
+    """WHEN the Domain of Analysis is updated
+    THEN the simulation_volume is updated accordingly
```
romane-m, line 149: .

tests/test_models.py
```diff
@@ -138,3 +142,51 @@ def test_build(simai_client):
...
+    model_conf.domain_of_analysis = doa
+
+    assert model_conf.simulation_volume.get("Z").get("type") == new_height.get("position")
+    assert model_conf.simulation_volume.get("Z").get("value") == new_height.get("value")
+    assert model_conf.simulation_volume.get("Z").get("length") == new_height.get("length")
+
+
+def test_get_doa(simai_client):
+    """WHEN the Domain of Analysis is retrieved
+    THEN the params of the axes match
```
romane-m, line 169: parameters ?
. (period)

tmpbeing: Hello Romane. Files in the `tests` folders are not public, these comments are only meant for the developers to understand what the test does. As such we don't follow the same conventions there, for example there's no need to specify the parameters as these functions are never called by humans, it's all automated.

romane-m: Hi Marc, I understood that, it was just a suggestion (hence the question mark) to replace "params" with "parameters" and add periods at the end of sentences to clean up comments :) But if it seems too superficial, @yias feel free to ignore them.

src/ansys/simai/core/data/models.py
```diff
@@ -21,9 +21,85 @@
...
+
+            # set the new DoA to the configuration
+
+            bld_conf.domain_of_analysis = new_doa
+
+    """
+
+    Length: DomainAxisDefinition = None
+    Width: DomainAxisDefinition = None
+    Height: DomainAxisDefinition = None
```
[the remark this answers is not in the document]
yias: I used upper case for the first letter so that it matches the definition in the UI; given that users would first use the UI and then the SDK, it would be straightforward for them to affiliate the parameters. Also, I wanted to avoid having `doa.length.length`. Anyways, I updated the names of the attributes. Resolving this comment and the other comments related to this one.

src/ansys/simai/core/data/models.py
yias: merge operator removed.

src/ansys/simai/core/data/models.py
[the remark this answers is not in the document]
yias: code updated to use `list` and `dict` instead of `typing.Dict` and `typing.List` respectively.

src/ansys/simai/core/data/models.py
yias: as we discussed, setting DoA in this fashion:
```py
new_doa = DomainOfAnalysis(
    height=("relative_to_min", 4.5, 4.1),
    length=("relative_to_center", 0.5, 15),
    width=("absolute", -1.5, 6.1),
)
is supported already. However, the class `DomainOfAnalysis` should be imported, which was not something very much aligned with the current examples, where all objects are pulled from core (e.g., `prjt = simai.projects.get()`). For that reason, the example showed this way of populating DoA. 

As discussed, importing classes as:
```py
from ansys.simai.core.data.models import DomainOfAnalysis
```
is acceptable, and, thus, I will update the example and I remove the method `ModelDirectory. doa_axis_definition`, and probably remove the method `ModelDirectory. domain_of_analysis`.

tests/test_models.py
yias: Thanks @romane-m for all the suggestions, they are very valuable :smile:

src/ansys/simai/core/data/models.py
yias: 1. regarding serialization/deserialization with pydantic, I replace here `type` to position but I could indeed use `BaseModel.model_dump(by_alias=True)` in `_set_doa_axis` and a `serialization_alias` for `position`, I haven't thought of it. Nevertheless, I don't think I could use any pydantic feature that converts "type" to "position" in `_get_doa_axis`.

2. `fld` is not optional. I check, however, if it is `None` and raise an exception to avoid user errors when partially setting DoA, e.g., the `Length` is defined but the `Width` and the `Height` are not. Adding `Optional` here would be misleading.

src/ansys/simai/core/data/models.py
yias: `pydantic.ValidationError` is final and it cannot be subclassed. Maybe not using pydantic overall would be a solution :/

src/ansys/simai/core/data/models.py
```diff
@@ -21,9 +21,85 @@
...
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Literal
+
+from pydantic import BaseModel, PositiveFloat, ValidationError, model_validator
 
 from ansys.simai.core.data.base import ComputableDataModel, Directory
+from ansys.simai.core.errors import InvalidArguments
+
+
+class DomainAxisDefinition(BaseModel):
```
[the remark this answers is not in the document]
yias: resolving as pydantic is not used after refactoring the solution.

src/ansys/simai/core/data/models.py
yias: resolving as the method `domain_of_analysis` is removed.

src/ansys/simai/core/data/models.py
```diff
@@ -120,3 +233,24 @@ def build(
...
+
+        """
+        try:
+            return DomainAxisDefinition(position=position, value=value, length=length)
+        except ValidationError as pydandic_exc:
+            raise InvalidArguments(pydandic_exc) from None
+
+    def domain_of_analysis(self):
+        """An empty :class:`DomainOfAnalysis <DomainOfAnalysis>` object."""
+        return DomainOfAnalysis()
```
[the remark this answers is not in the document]
yias: resolving as the methods `doa_axis_definition` and `domain_of_analysis` are removed from the class `ModelDirectory`.

src/ansys/simai/core/data/models.py
yias: the method `doa_axis_definition` is removed and the current implementation does not uses pydantic features. `InvalidArguments` exceptions are raised in cases of erroneous parameters of `DomainAxisDefinition`.

doc/source/api_reference/models.rst
```diff
@@ -14,6 +14,17 @@ the Global Coefficients and the Domain of Analysis, and its project id. The
...
+--------------------
+
+.. autoclass:: DomainAxisDefinition()
+
+
+DomainOfAnalysis
+--------------------
+
+.. autoclass:: DomainOfAnalysis()
+
```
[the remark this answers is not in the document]
yias: 1. I renamed to page to `Models`
2. Since the classes are still together in the same file, it's more coherent to have them all in the same page. I set the Directory and the Model first to be aligned with the rest of the API reference. Nevertheless, once more classes are introduced to cover the complete configuration (which is coming in the next PR), it'd make sense to put them in another file, and, thus, reflect that to the documentation.

src/ansys/simai/core/data/models.py
```diff
@@ -20,23 +20,238 @@
...
+            bld_conf.domain_of_analysis = new_doa
+
+    """
+
+    length: DomainAxisDefinition = None
+    width: DomainAxisDefinition = None
+    height: DomainAxisDefinition = None
+
+    def __post_init__(self):
+        if type(self.length) == tuple:
```
awoimbee, line 143:
```suggestion
        if isinstance(self.length, tuple):
```
To do everywhere, `type()` does not take inheritance into account

tests/test_models.py
```diff
@@ -134,7 +139,105 @@ def test_build(simai_client):
...
+
+    model_conf = ModelConfiguration(project_id=MODEL_RAW["project_id"], **MODEL_CONF_RAW)
+
+    new_height = {"position": "relative_to_center", "value": 0.5, "length": 15.2}
+
+    model_conf.domain_of_analysis.height = DomainAxisDefinition(**new_height)
+
+    assert model_conf._to_payload().get("simulation_volume").get("Z").get("type") == new_height.get(
+        "position"
+    )
```
awoimbee, line 158: nit but on top of being shorter, this also gives more meaningful errors (`KeyError: 'Z'` instead of `AttributeError: 'NoneType' object has no attribute 'get'`)
```suggestion
    assert model_conf._to_payload()["simulation_volume"]["Z"]["type"] == new_height["position"]
```

tests/test_models.py
```diff
@@ -134,7 +139,105 @@ def test_build(simai_client):
...
+    [
+        ({"position": "relative_to_center", "value": -0.5, "length": 15.2}),
+        ({"position": "relative_to_center", "value": 0.5, "length": -15.2}),
+    ],
+)
+def test_wrong_doa_axis_construction(doa_axis_raw):
+    """1: WHEN the parameter length of the DomainAxisDefinition is negative
+    2: WHEN the parameter value of the DomainAxisDefinition is negative AND the position is not absolute
+    THEN an InvalidArguments exception is raised.
+    """
```
awoimbee, line 194:
```suggestion
        # negative value with non-absolute position
        ({"position": "relative_to_center", "value": -0.5, "length": 15.2}),
        # negative length
        ({"position": "relative_to_center", "value": 0.5, "length": -15.2}),
    ],
)
def test_wrong_doa_axis_construction(doa_axis_raw):
    """WHEN the DomainAxisDefinition is invalid
    THEN an InvalidArguments exception is raised.
    """
```

tests/test_models.py
```diff
@@ -134,7 +139,105 @@ def test_build(simai_client):
...
+        DomainAxisDefinition(**doa_axis_raw)
+
+
+def test_wrong_doa_axis_length_in_assignment():
+    """WHEN a negative number is assigned to the parameter length of DomainAxisDefinition
+    THEN an InvalidArguments exception is raised.
+    """
+
+    doa_axis = DomainAxisDefinition("relative_to_center", 1.5, 15.2)
+
```
awoimbee, line 206: So many blank lines in there

awoimbee (approved):
LGTM, thanks

tests/test_models.py
yias: sure, commenting on the cases is good.

tmpbeing (approved):
:+1:

PR #46, opened 2024-04-25
fix: RecursionError on auth refresh

tmpbeing:
tentative fix for the RecursionError we've seen on some auth refresh. Also made the refresh timer slightly more aggressive to try and avoid the refresh being called when the token is already expired.


awoimbee (approved):
Ah I didn't think of that, nice

PR #49, opened 2024-04-30
feat: PointCloud and CustomVolumePointCloud

tmpbeing:


src/ansys/simai/core/api/point_cloud.py
```diff
@@ -0,0 +1,66 @@
...
+
+        Returns:
+            Tuple with the info of the input cloud and the URL to upload ID to use for further requests.
+        """
+        post_data = {"name": name, "file_extension": extension}
+        response = self._post(f"geometries/{geometry_id}/point-cloud", json=post_data)
+        return (response["point_cloud"], response["upload_id"])
+
+    def complete_point_cloud_upload(
+        self, point_cloud_id: str, upload_id: str, parts: List[Dict[str, Any]]
```
yias, line 45: nit: it'd be nicer if we'd use `list`, `dict`, and `tuple` instead the `typing` aliases `List`, `Dict` and `Tuple`, since they are deprecated.

src/ansys/simai/core/api/point_cloud.py
```diff
@@ -0,0 +1,66 @@
...
+            geometry_id: ID of the geometry to assign the input cloud to.
+            name: Name to give to the geometry.
+            extension: Extension to give to the file.
+
+        Returns:
+            Tuple with the info of the input cloud and the URL to upload ID to use for further requests.
+        """
+        post_data = {"name": name, "file_extension": extension}
+        response = self._post(f"geometries/{geometry_id}/point-cloud", json=post_data)
+        return (response["point_cloud"], response["upload_id"])
```
yias, line 42: returning part of the response contradicts the concept that exists in the rest of the `ApiClientMixin` classes where the full response is being return. To have a cleaner implementation, it would make sense to follow the same principle here, and let `upload_point_cloud` manage the response.

src/ansys/simai/core/data/geometries.py
```diff
@@ -170,7 +182,9 @@ def get_predictions(self) -> List["Prediction"]:
         return [self._client.predictions._model_from(pred_data) for pred_data in predictions_data]
 
     def download(
-        self, file: Optional[File] = None, monitor_callback: Optional[MonitorCallback] = None
+        self,
+        file: Optional[File] = None,
+        monitor_callback: Optional[MonitorCallback] = None,
```
yias, line 187: is this just a diff in formatting? strange 😕

yias (commented):
I'm wondering if we need unit tests for the new functionality

src/ansys/simai/core/api/point_cloud.py
tmpbeing: We can't use them since we supposedly still support python 3.9 (supposedly because we actually broke it in the latest release) which doesn't support them.

awoimbee (approved):
LGTM but as Iason said, please add tests

src/ansys/simai/core/api/point_cloud.py
```diff
@@ -0,0 +1,66 @@
...
+           parts: List of the uploaded file parts
+        """
+        self._post(
+            f"point-clouds/{point_cloud_id}/complete",
+            json={"upload_id": upload_id, "parts": parts},
+            return_json=False,
+        )
+
+    def delete_point_cloud(self, point_cloud_id: str):
+        """Delete the given input cloud.
```
romane-m, line 61:
```suggestion
        """Delete the specified point cloud file.
```

src/ansys/simai/core/api/point_cloud.py
```diff
@@ -0,0 +1,66 @@
...
+# SOFTWARE.
+
+from typing import Any, Dict, List, Tuple
+
+from ansys.simai.core.api.mixin import ApiClientMixin
+
+
+class PointCloudClientMixin(ApiClientMixin):
+    def create_point_cloud(self, geometry_id: str, name: str, extension: str) -> Tuple[Dict, str]:
+        """Create an input cloud without pushing the data.
```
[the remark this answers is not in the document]
romane-m: (switch to make everywhere it is relevant)

src/ansys/simai/core/api/point_cloud.py
```diff
@@ -0,0 +1,66 @@
...
+
+        Args:
+            geometry_id: ID of the geometry to assign the input cloud to.
+            name: Name to give to the geometry.
+            extension: Extension to give to the file.
+
+        Returns:
+            Tuple with the info of the input cloud and the URL to upload ID to use for further requests.
```
romane-m, line 38:
```suggestion
           Tuple containing the point cloud's information and the URL to use for subsequent requests.
```

src/ansys/simai/core/data/post_processings.py
```diff
@@ -503,6 +527,47 @@ def volume_vtu(self, run: bool = True) -> Optional[VolumeVTU]:
         """
         return self._get_or_run(VolumeVTU, {}, run)
 
+    def custom_volume_point_cloud(self, run: bool = True) -> Optional[CustomVolumePointCloud]:
+        """Compute or get the result of a predicted volume depending on the geometry's point cloud.
```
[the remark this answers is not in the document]
tmpbeing: I used similar terminology as the other post processing, if it's an issue, feel free to open a PR where you change all of them but for now I'll keep it consistent.

src/ansys/simai/core/data/post_processings.py
```diff
@@ -503,6 +527,47 @@ def volume_vtu(self, run: bool = True) -> Optional[VolumeVTU]:
...
+
+        This is a non-blocking method. It returns the :class:`CustomVolumePointCloud`
```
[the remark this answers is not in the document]
tmpbeing: Again, the suggestion is not bad but I think should be bad of a broader PR changing all of them.

src/ansys/simai/core/data/post_processings.py
```diff
@@ -263,6 +269,24 @@ class SurfaceVTP(_PostProcessingVTKExport):
...
 
+class CustomVolumePointCloud(PostProcessing):
+    """Representation of a CustomVolumePointCloud post-processing.
+
+    This class is generated through the :meth:`~PredictionPostProcessings.custom_volume_point_cloud()` method.
+    """
+
+    @property
+    def data(self) -> DownloadableResult:
+        """:class:`DownloadableResult` object that allows access to the custom volume VTP
```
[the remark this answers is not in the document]
tmpbeing: Same comment (bigger PR where we change all of them)

PR #50, opened 2024-05-03
Enable download to directory

yias:
## Description

This PR enables to download files in to a directory.

story: [sc-6135]

## Usage

```py
simai_client.design_of_experiments.download(dir=tmp_dir, format="xlsx")
```

## Tests

- unit testing
- smoke test against Dev

### DIFF
--- a/src/ansys/simai/core/api/design_of_experiments.py
+++ b/src/ansys/simai/core/api/design_of_experiments.py
@@ -36,7 +36,7 @@ class DesignOfExperimentsMixin(ApiClientMixin):
     def download_design_of_experiments(
         self,
         file: Optional[File],
-        dir: Optional[str | Path],
+        dir: Optional[Union[str, Path]],
         format: str,
         workspace_id: str,
     ) -> Union[None, BinaryIO]:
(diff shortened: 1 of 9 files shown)


src/ansys/simai/core/api/design_of_experiments.py
```diff
@@ -35,13 +36,15 @@ class DesignOfExperimentsMixin(ApiClientMixin):
     def download_design_of_experiments(
         self,
         file: Optional[File],
+        dir: Optional[str | Path],
```
awoimbee, line 39: Well, seeme like the CI workflows we got from the pyansys team doesn't actually run the tests, 'cause that's not valid python 3.9.
https://github.com/ansys/pysimai/actions/runs/8941539158/job/24562116546?pr=50

src/ansys/simai/core/api/mixin.py
```diff
@@ -154,6 +156,11 @@ def download_file(
             None if a file is provided or a ``BytesIO`` object with the file's
             content otherwise.
         """
+        if dir is not None and file is None:
```
awoimbee, line 159: We could just `if file.is_dir():`, it would be less weird to use than having a `dir` arg that only works when `file` is `None`.

yias: To me, this makes more sense to have a dedicate argument for dir, for having it explicit.

yias:
closed for recasting

PR #53, opened 2024-05-27
bump: wakepy 0.8.0

tmpbeing:
Should fix some issues some users had when wakepy failed to lock.

### DIFF
--- a/pdm.lock
+++ b/pdm.lock
@@ -5,64 +5,65 @@
 groups = ["default", "doc", "linting", "test"]
 strategy = ["cross_platform", "inherit_metadata"]
 lock_version = "4.4.1"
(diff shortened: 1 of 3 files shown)


awoimbee (approved):

yias (commented):
`on_fail` is introduced in `wakepy` version `8.0`, which makes it reasonable to update the `wakepy` version to `8.0`. But, since we set it as warning, we avoid raising an error. Was throwing an error the blocker?

awoimbee:
> Was throwing an error the blocker?

See [sc-18416](https://app.shortcut.com/simai/story/18416) and the related support ticket.

yias (approved):
ok, it makes sense now.

PR #54, opened 2024-05-28
fix some crashes caused by client. _current_workspace

awoimbee:
[\[sc-18863\]](https://app.shortcut.com/simai/story/18863)

---

I really think we should completely remove the `current_workspace` thing, it's still used in places like `client.post_processings.info` where it's a confusing mess that is very likely to either:
* return info for an unrelated workspace that the user doesn't care about
* crash (`InvalidClientStateError`)


src/ansys/simai/core/data/types.py
```diff
@@ -236,9 +252,9 @@ def unpack_named_file(
 
 
 def get_id_from_identifiable(
-    identifiable: Optional[Identifiable] = None,
+    identifiable: "Optional[Identifiable[D]]" = None,
```
tmpbeing, line 255:
```suggestion
    identifiable: Optional[Identifiable["D"]] = None,
```

src/ansys/simai/core/data/types.py
```diff
@@ -236,9 +252,9 @@ def unpack_named_file(
...
     required: bool = True,
-    default: Optional[Identifiable] = None,
+    default: "Optional[Identifiable[D]]" = None,
```
tmpbeing, line 257:
```suggestion
    default: Optional[Identifiable["D"]] = None,
```

src/ansys/simai/core/data/workspaces.py
```diff
@@ -20,17 +20,21 @@
...
+import warnings
 from pprint import pformat
-from typing import Any, BinaryIO, Dict, List, Optional, Union
+from typing import TYPE_CHECKING, Dict, List, Optional, Union
 
 from ansys.simai.core.data.base import DataModel, Directory
 from ansys.simai.core.data.types import File, Identifiable, get_id_from_identifiable
 
+if TYPE_CHECKING:
+    from typing import Any, BinaryIO
```
tmpbeing, line 31: I don't understand why the typing changes in that file and why are those in `if TYPE_CHECKING` ?
Not a fan of typing in quotes, will make it more annoying to refactor once we can switch to `dict`/`list`...

awoimbee: Yeah I just forgot those after testing stuff

PR #57, opened 2024-06-04
Allow assign subset to None; Retire: Ignored and Validation

kliment-slice:
[sc-19132]
[sc-19134]

### DIFF
--- a/src/ansys/simai/core/data/training_data.py
+++ b/src/ansys/simai/core/data/training_data.py
@@ -84,28 +84,33 @@ class TrainingData(ComputableDataModel):
                 the :class:`~.projects.Project` object for, or its ID.
 
         Returns:
-            SubsetEnum of the subset that the training data belongs to in the given project.
-                (e.g. <SubsetEnum.VALIDATION: 'Validation'>)
+            None if subset is unassigned or SubsetEnum of the subset that the training data belongs to in the given project.
+                (e.g. <SubsetEnum.TEST: 'Test'>)
         """
         project_id = get_id_from_identifiable(project, default=self._client._current_project)
         subset_value = self._client._api.get_training_data_subset(project_id, self.id).get("subset")
         return SubsetEnum(subset_value) if subset_value else None
 
-    def assign_subset(self, project: Identifiable["Project"], subset: SubsetEnum) -> None:
+    def assign_subset(self, project: Identifiable["Project"], subset: Optional[SubsetEnum]) -> None:
         """Assign the training data subset in relation to a given project.
 
         Args:
             project: ID or :class:`model <.projects.Project>`
-            subset: SubsetEnum attribute (e.g. SubsetEnum.TRAINING) or string value (e.g. "Training").
-                Available options: (Training, Validation, Test, Ignored)
+            subset: None to unassign or SubsetEnum attribute (e.g. SubsetEnum.TRAINING) or string value (e.g. "Training").
+                Available options: (Training, Test)
 
         Returns:
             None
         """
-        if subset not in SubsetEnum.__members__.values():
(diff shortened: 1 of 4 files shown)


src/ansys/simai/core/data/training_data.py
```diff
@@ -84,28 +84,33 @@ def get_subset(self, project: Identifiable["Project"]) -> Optional[SubsetEnum]:
                 the :class:`~.projects.Project` object for, or its ID.
 
         Returns:
-            SubsetEnum of the subset that the training data belongs to in the given project.
-                (e.g. <SubsetEnum.VALIDATION: 'Validation'>)
+            None if subset is unassigned or SubsetEnum of the subset that the training data belongs to in the given project.
```
awoimbee, line 87:
```suggestion
            If the subset is unassigned, None; otherwise, the SubsetEnum of the subset to which the training data belongs in the given project.
```

src/ansys/simai/core/data/training_data.py
```diff
@@ -84,28 +84,33 @@ def get_subset(self, project: Identifiable["Project"]) -> Optional[SubsetEnum]:
...
 
-    def assign_subset(self, project: Identifiable["Project"], subset: SubsetEnum) -> None:
+    def assign_subset(self, project: Identifiable["Project"], subset: Optional[SubsetEnum]) -> None:
         """Assign the training data subset in relation to a given project.
 
         Args:
             project: ID or :class:`model <.projects.Project>`
-            subset: SubsetEnum attribute (e.g. SubsetEnum.TRAINING) or string value (e.g. "Training").
-                Available options: (Training, Validation, Test, Ignored)
+            subset: None to unassign or SubsetEnum attribute (e.g. SubsetEnum.TRAINING) or string value (e.g. "Training").
```
awoimbee, line 99: nit:
```suggestion
            subset: SubsetEnum attribute (e.g. SubsetEnum.TRAINING) or string value (e.g. "Training") or None to unassign.
```

src/ansys/simai/core/data/training_data.py
```diff
@@ -84,28 +84,33 @@ def get_subset(self, project: Identifiable["Project"]) -> Optional[SubsetEnum]:
...
-        if subset not in SubsetEnum.__members__.values():
-            raise InvalidArguments("Must be one of: Ignored, Training, Test, Validation.")
+        if subset is not None and subset not in SubsetEnum.__members__.values():
+            raise InvalidArguments("Must be None or one of: 'Training', 'Test'.")
+
         project_id = get_id_from_identifiable(project, default=self._client._current_project)
-        self._client._api.put_training_data_subset(project_id, self.id, subset)
+        if subset is None:
+            self._client._api.remove_training_data_from_project(self.id, project_id)
+            self._client._api.add_training_data_to_project(self.id, project_id)
```
awoimbee, line 111: That's a hack, maybe `put_training_data_subset` should allow `None` so we don't need this

awoimbee (commented):
> db migration for

Why are we talking about DB migrations in SDK changes ? Especially as everything here is public.

src/ansys/simai/core/data/training_data.py
tmpbeing:
```suggestion
            The ``SubsetEnum`` of the subset to which the training belongs to if any, ``None`` otherwise.
```
If you're less lazy than me, you can make SubsetEnum link to the class

yias (approved):
my small warning here is that this implementation is not backwards compatible; since `Validation` and `Ignored` are removed, scripts that use them already (if any) would break. I trust that this is already discussed and it represents a corner case 😉 

Beyond that, it looks fine to me 😃

tmpbeing (approved):

awoimbee (approved):
Nice

PR #68, opened 2024-07-08
Doc: Improvements in training_data.py

romane-m:
[Shortcut ticket](https://app.shortcut.com/simai/story/23032/sdk-edits-for-tain-test-val-disclaimer)
This PR contains two edits made in the training_data.py file:

- One edit to correct a rendering issue in the current documentation.
- The addition of a paragraph to improve the definition of the assign_subset method's arguments.

### DIFF
--- a/src/ansys/simai/core/data/training_data.py
+++ b/src/ansys/simai/core/data/training_data.py
@@ -87,8 +87,9 @@ class TrainingData(ComputableDataModel):
                 the :class:`~.projects.Project` object for, or its ID.
 
         Returns:
-            The :obj:`~ansys.simai.core.data.types.SubsetEnum` of the subset to which the :class:`TrainingData` belongs to if any, ``None`` otherwise.
-                (e.g. <SubsetEnum.TEST: 'Test'>)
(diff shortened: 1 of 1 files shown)


src/ansys/simai/core/data/training_data.py
```diff
@@ -97,13 +98,28 @@ def get_subset(self, project: Identifiable["Project"]) -> Optional[SubsetEnum]:
     def assign_subset(
         self, project: Identifiable["Project"], subset: Optional[Union[SubsetEnum, str]]
     ) -> None:
-        """Assign the training data subset in relation to a given project.
+        """Assign the training data to a subset in relation to a given project.
 
         Args:
             project: ID or :class:`model <.projects.Project>`
             subset: SubsetEnum attribute (e.g. SubsetEnum.TRAINING) or string value (e.g. "Training") or None to unassign.
```
romane-m, line 105: General question: Could we only give the enum option instead of both enum and string as the string option is error-prone. What flexibility does the string offer the user?

tmpbeing: We could. But as a developer typing a string is more convenient than importing and using an Enum albeit more error prone.

PR #71, opened 2024-07-22
chore: remove compute argument from td upload

tmpbeing:
This option could cause confusion and it has been decided to remove it

### DIFF
--- a/src/ansys/simai/core/data/training_data.py
+++ b/src/ansys/simai/core/data/training_data.py
@@ -164,21 +164,18 @@ class TrainingData(ComputableDataModel):
     def upload_folder(
         self,
         folder_path: Path,
-        compute: bool = True,
     ) -> List["TrainingDataPart"]:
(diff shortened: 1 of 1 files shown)


src/ansys/simai/core/data/training_data.py
```diff
@@ -164,21 +164,16 @@ def upload_part(
     def upload_folder(
         self,
         folder_path: Path,
-        compute: bool = True,
     ) -> List["TrainingDataPart"]:
         """Upload all the parts contained in a folder to a :class:`~ansys.simai.core.data.training_data.TrainingData` instance.
 
-        This method automatically requests computation of the training data
```
camzeech, line 171: One could leave bits of this sentence, WDYT? For instance, by replacing with
"Upon upload completion, SimAI will extract data from each part."

camzeech: @tmpbeing

awoimbee (approved):
Good thing I stopped using `upload_folder` in the e2e tests :grindff:

PR #73, opened 2024-07-26
PostProcessInput class

kliment-slice:
[sc-18819]


tests/test_models.py
```diff
@@ -475,3 +476,52 @@ def test_sse_event_handler(simai_client, model_factory):
...
+    MODEL_CONFIG["fields"]["surface_pp_input"] = [
+        {"keys": None, "name": "Temperature_1", "unit": None, "format": "value", "location": "cell"}
+    ]
+
+    MODEL_REQUEST = MODEL_RAW.copy()
+    MODEL_REQUEST["configuration"] = MODEL_CONFIG
+
+    surface_pp_input = PostProcessInput(surface=["Temperature_1"])
+
+    config_with_pp_input = ModelConfiguration(
```
yias, line 513: since `MODEL_CONFIG` already has a key `surface_pp_input`, this `surface_pp_input` you set here will be overwritten from what it exists in `MODEL_CONFIG`. Are you sure this is what you'd like to test here? 

To me, it would be more preferable to set here the configuration using the individual arguments, as in `test_build_with_new_config` including `surface_pp_input`, for letting the constructor of ModelConfig to consume it, and then compare the result against the actual `surface_pp_input`

I mean:

```py
    model_input = ModelInput(surface=[], boundary_conditions=["Vx"])
    model_output = ModelOutput(
        surface=["Pressure", "WallShearStress_0"], volume=["Velocity_0", "Pressure"]
    )
    global_coefficients = [("max(Pressure)", "maxpress")]
    simulation_volume = {
        "X": {"length": 300.0, "type": "relative_to_min", "value": 15.0},
        "Y": {"length": 80.0, "type": "absolute", "value": -80},
        "Z": {"length": 40.0, "type": "absolute", "value": -20.0},
    }

    config_with_pp_input = ModelConfiguration(
        project=project,
        build_preset="debug",
        continuous=False,
        input=model_input,
        output=model_output,
        global_coefficients=global_coefficients,
        simulation_volume=simulation_volume,
        surface_pp_input=surface_pp_input,
    )

    responses.add(
        responses.POST,
        f"https://test.test/projects/{MODEL_REQUEST['project_id']}/model",
        json=MODEL_REQUEST,
        status=200,
    )

    build_model: Model = simai_client.models.build(config_with_pp_input)
    assert build_model.configuration.surface_pp_input.surface == surface_pp_input.surface
```

yias (commented):
the implementation looks good, just a small change in the test to make it more concrete ;)

src/ansys/simai/core/data/model_configuration.py
```diff
@@ -237,6 +250,9 @@ class ModelConfiguration:
                 surface=["Pressure", "WallShearStress_0"], volume=["Velocity_0", "Pressure"]
             )
 
+            # Define the surface post-processing input
+            surface_pp_input = PostProcessInput(surface=["Temperature_1"])
```
awoimbee, line 254: On one hand `PostProcessInput` is designed so it can receive other fields than `surface`, but on the other hand, the argument is called `surface_pp_input` (and eg not `pp_input`).

tests/test_models.py
```diff
@@ -475,3 +476,52 @@ def test_sse_event_handler(simai_client, model_factory):
...
+        responses.GET,
+        f"https://test.test/projects/{MODEL_RAW['project_id']}",
+        json=raw_project,
+        status=200,
+    )
+
+    project: Project = simai_client._project_directory._model_from(raw_project)
+    project.verify_gc_formula = Mock()
+
+    MODEL_CONFIG = MODEL_CONF_RAW.copy()
```
awoimbee, line 503: uppercase is generally used for global constants, this looks like a local variable

tests/test_models.py
```diff
@@ -475,3 +476,52 @@ def test_sse_event_handler(simai_client, model_factory):
...
+    MODEL_CONFIG["fields"]["surface_pp_input"] = [
```
awoimbee, line 504: You didn't do a deep copy so you *might* be mutating `MODEL_CONF_RAW` in here

src/ansys/simai/core/data/model_configuration.py
kliment-slice: arg renamed to `pp_input`; field `surface_pp_input` remains in raw model config dict!

tests/test_models.py
kliment-slice: added more assertions for completeness; `pp_input` remains in mock request json otherwise response would be empty

awoimbee (approved):

PR #74, opened 2024-07-31
fix: optimization failing on Mode.success

tmpbeing:
Seems success() is gone in the latest versions of wakepy


awoimbee (approved):
> Seems running() is gone

hope not !

---

btw we need tests

kliment-slice (approved):

PR #77, opened 2024-08-08
fix: NaN and inf conversion

tmpbeing:
Less naive implementation that we can reuse in the future

### DIFF
--- a/src/ansys/simai/core/data/global_coefficients_requests.py
+++ b/src/ansys/simai/core/data/global_coefficients_requests.py
@@ -179,7 +179,7 @@ class ComputeGlobalCoefficient(GlobalCoefficientRequest):
             self._set_is_pending()
         elif state == "successful":
             logger.debug(f"{self._classname} id {self.id} set status successful")
-            self._result = {k: float(v) for k, v in data.get("result", {}).get("value", {}).items()}
+            self._result = float(data.get("result", {}).get("value"))
             self._set_is_over()
         elif state in ERROR_STATES:
             error_message = f"Computation of global coefficient {target.get('formula')} failed with {data.get('reason', 'UNKNOWN ERROR')}"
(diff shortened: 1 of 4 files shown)


src/ansys/simai/core/utils/numerical.py
```diff
@@ -113,3 +113,13 @@ def validate_tolerance_parameter(tolerance: Optional[Number]):
...
+
+
+# Recursively go through dict to cast values to float, so that we support NaN and inf
+def cast_values_to_float(data: Any):
+    if isinstance(data, dict):
+        return {k: cast_values_to_float(v) for k, v in data.items()}
+    elif isinstance(data, str) and data.lower() in ["nan", "inf", "-inf"]:
+        return float(data)
+    else:
+        return data
```
awoimbee, line 125: IMO We don't want to run this function on any random dict (and convert a string that should be kept string into float). This is reserved for places where the value should be float, we can still handle invalid value, but we need to handle it as such.
```suggestion
    try:
        return float(data)
    except ValueError:
        logger.waning(f"received invalid float value: {data!r}")
        return float("nan")
```

src/ansys/simai/core/data/global_coefficients_requests.py
```diff
@@ -179,7 +179,7 @@ def _handle_job_sse_event(self, data):
             self._set_is_pending()
         elif state == "successful":
             logger.debug(f"{self._classname} id {self.id} set status successful")
-            self._result = {k: float(v) for k, v in data.get("result", {}).get("value", {}).items()}
+            self._result = float(data.get("result", {}).get("value"))
```
awoimbee, line 182:
```suggestion
            self._result = cast_values_to_float(data.get("result", {}).get("value"))
```

awoimbee (approved):

src/ansys/simai/core/utils/numerical.py
tmpbeing: Hmm I'm fine with that but then I think we have to reraise instead of returning nan.

PR #81, opened 2024-09-09
tweak: small typing improvement

tmpbeing:
https://mypy.readthedocs.io/en/stable/literal_types.html#literal-types

### DIFF
--- a/src/ansys/simai/core/utils/requests.py
+++ b/src/ansys/simai/core/utils/requests.py
@@ -22,10 +22,11 @@
 
 import logging
 from json.decoder import JSONDecodeError
+from typing import Literal, overload
 
 import requests
 
-from ansys.simai.core.data.types import APIResponse
+from ansys.simai.core.data.types import JSON, APIResponse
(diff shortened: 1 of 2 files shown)


yias (approved):
LGTM

src/ansys/simai/core/utils/requests.py
```diff
@@ -76,6 +77,21 @@ def handle_http_errors(response: requests.Response) -> None:
...
+
+
+@overload
+def handle_response(response: requests.Response, return_json: Literal[False]) -> requests.Response:
+    ...
+
+
+@overload
+def handle_response(response: requests.Response, return_json: bool) -> APIResponse:
+    ...
```
awoimbee, line 92:
```suggestion
```

You don't need this overload since it's exactly the same as the function signature. Also all the cases are covered by the other overloads.

awoimbee (approved):

src/ansys/simai/core/utils/requests.py
tmpbeing: I know it looks dumb but the documentation says you do.

PR #82, opened 2024-09-11
fix: error on accessing global coeffcients data

tmpbeing:


yias (approved):

tests/test_post_processings_results.py
```diff
@@ -62,14 +62,18 @@ def test_post_processing_result_global_coefficients(simai_client):
                             "X": 4.441544089820862,
                             "Y": 0.32894839148911376,
                             "Z": 14.442021446988662,
-                        }
+                        },
+                        "type": "base",
+                        "unit": {"N": 1},
```
nuttincka, line 67: Not a big deal but is "unit" useful here? As far I know it is never used.

tmpbeing: I put it in the test because that's the current content of the global coefficients. Whether it should be there or not is not a question for me.

src/ansys/simai/core/data/post_processings.py
```diff
@@ -187,7 +187,10 @@ def data(self) -> Dict[str, List]:
         self.wait()
 
         results = self._get_results()
-        return cast_values_to_float(results["data"]["values"])
+        return {
+            k: {**v, "data": cast_values_to_float(v["data"])}
+            for k, v in results["data"]["values"].items()
+        }
```
awoimbee, line 193: that's hard to read !

awoimbee (approved):
thanks !

PR #84, opened 2024-09-12
build preset to take only supported values

yias:
## Description

This PR forces the value of `build_preset` in model configuration to be one of the supported values. 

When the user sets a non supported values in `build_preset`, an `InvalidArgument` error gets raised.

Story: [sc-24065]

### DIFF
--- a/src/ansys/simai/core/data/model_configuration.py
+++ b/src/ansys/simai/core/data/model_configuration.py
@@ -217,11 +217,11 @@ class ModelConfiguration:
 
                     | *debug*: < 30 min, only 4 dat
 
-                    | *short*: < 24 hours
+                    | *1_day*: < 24 hours
 
-                    | *default*: < 2 days, default value.
+                    | *2_days*: < 2 days, default value.
 
-                    | *long*: < 1 week
+                    | *7_days*: < 1 week
         continuous: indicates if continuous learning is enabled. Default is False.
         input: the inputs of the model.
         output: the outputs of the model.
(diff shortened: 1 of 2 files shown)


src/ansys/simai/core/data/model_configuration.py
```diff
@@ -313,11 +314,26 @@ def __get_gc(self):
...
+            )
+
+    def __set_build_preset(self, val: str):
+        self.__validate_build_preset(val)
+        self.__dict__["build_preset"] = val
+
+    def __get_build_preset(self):
+        return self.__dict__.get("build_preset")
+
+    build_preset = property(__get_build_preset, __set_build_preset)
```
awoimbee, line 330: instead of:
```python
def __validate_build_preset(): ...
def __set_build_preset(): ...
def __get_build_preset(self): ...

build_preset = property(__get_build_preset, __set_build_preset)
[...]
del __set_build_preset, __get_build_preset
```

it looks like we could have just:
```python
def __get_build_preset(self): ...
def __set_build_preset(): ...

build_preset = property(__get_build_preset, __set_build_preset)
```

awoimbee (approved):
LGTM

I have to say I still think all of this would be so much easier with pydantic

src/ansys/simai/core/data/model_configuration.py
yias: I actually prefer to put these helper functions out of the scope:
1. I like the symantics of it, i.e., since they are deleted methods, it means they are only there to serve a specific purpose
2. the user cannot call them, e.g.,:
```py
bld_conf = ModelConfiguration(
    project=my_project,
    build_preset="debug",
)
```
raises the error:
```sh
AttributeError: 'ModelConfiguration' object has no attribute '__get_build_preset'
```

yias: 1. thanks pointing out that `__` attributes are hidden from the user. My argument on the semantics is still valid though.
2. although `__validate_build_preset` is not currently used elsewhere, having a dedicate method for it segments better the responsibilities of the methods conceptually.  

Hope this answers everything.

PR #85, opened 2024-09-13
refactor auth tokens retrieval/refresh  process

awoimbee:
* make pysimai kinda multiprocess safe:
  * tokens are stored on disk (not in RAM)
  * each process try to renew tokens at a different time
* invalid refresh token leads to reauth instead of crash

Fixes:
* issues caused by using multiple processes (eg `multiprocessing`)
* invalid refresh token (eg invalidated by the auth server) now triggers reauth instead of crash

### DIFF
--- a/src/ansys/simai/core/utils/auth.py
+++ b/src/ansys/simai/core/utils/auth.py
@@ -198,7 +198,7 @@ class _AuthTokensRetriever:
             with open(self.cache_file_path + "~", "w") as f:
                 f.write(auth.model_dump_json())
             # rename is atomic
-            os.rename(self.cache_file_path + "~", self.cache_file_path)
+            os.replace(self.cache_file_path + "~", self.cache_file_path)
         self._schedule_auth_refresh(auth.refresh_expires_in)
         return auth
 
--- a/src/ansys/simai/core/utils/auth.py
+++ b/src/ansys/simai/core/utils/auth.py
@@ -62,7 +62,7 @@ class _AuthTokens(BaseModel):
     def expires_in_to_datetime(cls, data: Any) -> dict:
         assert isinstance(data, dict)
         if "expiration" not in data:
-            # We want and store "expiration" but API responses contain "expires_in"
+            # We want to store "expiration" but API responses contains "expires_in"
             now = datetime.now(timezone.utc)
             data["expiration"] = now + timedelta(seconds=int(data["expires_in"]))
             data["refresh_expiration"] = now + timedelta(seconds=int(data["refresh_expires_in"]))
@@ -86,7 +86,7 @@ class _AuthTokens(BaseModel):
 
 
 class _AuthTokensRetriever:
-    """Class who's sole purpose is ``get_tokens()``.
+    """Retrieve tokens via ``get_tokens()``.
     It handles caching and the various auth token sources.
     """
 
@@ -170,7 +170,7 @@ class _AuthTokensRetriever:
     def _schedule_auth_refresh(self, refresh_expires_in: float):
         """Schedule authentication refresh to avoids refresh token expiring if the client is idle for a long time."""
         if refresh_expires_in <= self.REFRESH_BUFFER:
-            # if the refresh token expires this soon it means we are reaching max sso session lifespan
+            # Skip scheduling: refresh token too close to expiration (max SSO session nearly reached)
             return
         self.refresh_timer.cancel()
         self.refresh_timer = threading.Timer(
--- a/src/ansys/simai/core/utils/auth.py
+++ b/src/ansys/simai/core/utils/auth.py
@@ -179,13 +179,13 @@ class _AuthTokensRetriever:
         self.refresh_timer.daemon = True
         self.refresh_timer.start()
 
-    def get_tokens(self, fresh_token=False) -> _AuthTokens:
+    def get_tokens(self, force_refresh=False) -> _AuthTokens:
         cache_is_outdated = False
         auth = self._get_token_from_cache()
         if auth:
             if auth.is_refresh_token_expired():
                 auth = None
-            elif fresh_token or auth.is_token_expired():
+            elif force_refresh or auth.is_token_expired():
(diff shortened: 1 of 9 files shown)


src/ansys/simai/core/utils/auth.py
```diff
@@ -42,75 +45,101 @@
...
-    def __init__(self, *args, **kwargs) -> None:
-        super().__init__(*args, **kwargs)
-        if self.expiration is None:
-            now = datetime.utcnow()
-            self.expiration = now + timedelta(seconds=self.expires_in)
-            self.refresh_expiration = now + timedelta(seconds=self.refresh_expires_in)
-        if self.cache_file is not None:
-            with open(self.cache_file, "w") as f:
-                f.write(self.model_dump_json())
+    # ...  (unused fields removed)
```
yias, line 58: It looks like this line could be removed too :)

src/ansys/simai/core/utils/auth.py
```diff
@@ -42,75 +45,101 @@
...
 
+    @model_validator(mode="before")
     @classmethod
-    def from_cache(cls, file_path: Optional[str]) -> "Optional[_AuthTokens]":
-        if file_path is None:
-            return None
+    def expires_in_to_datetime(cls, data: Any) -> dict:
+        assert isinstance(data, dict)
```
yias, line 63: if we assert that `data` is of type `dict`, would it make sense its type hint to be `dict` as well (instead of Any)?

src/ansys/simai/core/utils/auth.py
```diff
@@ -119,54 +148,59 @@ def from_request_device_auth(
...
+        """Schedule authentication refresh to avoids refresh token expiring if the client is idle for a long time."""
+        if refresh_expires_in <= self.REFRESH_BUFFER:
+            # if the refresh token expires this soon it means we are reaching max sso session lifespan
+            return
+        self.refresh_timer.cancel()
+        self.refresh_timer = threading.Timer(
+            refresh_expires_in - self.REFRESH_BUFFER, self.get_tokens
+        )
+        self.refresh_timer.daemon = True
+        self.refresh_timer.start()
```
yias, line 180: it looks like there might be a chance for a race condition here; 

if `refresh_expires_in - self.REFRESH_BUFFER` is small enough, let's say _0.1s_, and the execution of `get_tokens` lasts longer than _0.1s_, more than one instances of `get_tokens` would run in parallel. Since `get_tokens` reads/writes from/to file, this could create problems. WDYT?

src/ansys/simai/core/utils/auth.py
```diff
@@ -119,54 +148,59 @@ def from_request_device_auth(
...
+
+    def get_tokens(self, fresh_token=False) -> _AuthTokens:
```
yias, line 182: nit: `fresh_token` confused me a bit. Maybe `refresh_token` or `flush_token` would be a bit more clear 🤔

src/ansys/simai/core/utils/auth.py
```diff
@@ -42,75 +45,101 @@
...
 
+    @model_validator(mode="before")
     @classmethod
-    def from_cache(cls, file_path: Optional[str]) -> "Optional[_AuthTokens]":
-        if file_path is None:
-            return None
+    def expires_in_to_datetime(cls, data: Any) -> dict:
+        assert isinstance(data, dict)
+        if "expiration" not in data:
+            # We want and store "expiration" but API responses contain "expires_in"
```
yias, line 65: We want __to__ store "expiration" ... ? The __and__ doesn't make much sense here

src/ansys/simai/core/utils/auth.py
```diff
@@ -42,75 +45,101 @@
...
+        return (self.expiration - datetime.now(timezone.utc)).total_seconds()
+
+    @computed_field
+    @property
+    def refresh_expires_in(self) -> float:
+        return (self.refresh_expiration - datetime.now(timezone.utc)).total_seconds()
+
+
+class _AuthTokensRetriever:
+    """Class who's sole purpose is ``get_tokens()``.
```
yias, line 89:
```suggestion
    """Retrieve tokens.
```
or, at least, change __who's__ to __which__

src/ansys/simai/core/utils/auth.py
```diff
@@ -119,54 +148,59 @@ def from_request_device_auth(
...
-        auth_cache_path = str(get_cache_dir() / f"tokens-{auth_cache_hash}.json")
-        auth = _AuthTokens.from_cache(auth_cache_path)
+        except (requests.exceptions.ConnectionError, ApiClientError) as e:
+            logger.error(f"Could not refresh authentication tokens: {e}")
+            return None
+
+    def _schedule_auth_refresh(self, refresh_expires_in: float):
+        """Schedule authentication refresh to avoids refresh token expiring if the client is idle for a long time."""
+        if refresh_expires_in <= self.REFRESH_BUFFER:
+            # if the refresh token expires this soon it means we are reaching max sso session lifespan
```
yias, line 173:
```suggestion
            # if the refresh token expires soon, it means we are reaching max sso session lifespan
```

yias (commented):
Very good, thanks. Just a clarification on parallelism from me 😃

src/ansys/simai/core/utils/auth.py
awoimbee: The class is described as `Represents the OIDC tokens received from the auth server.`, I think it's important to note that it ignores some fields

src/ansys/simai/core/utils/auth.py
awoimbee: Yes I just reduced the odds but not eradicated the race condition to avoid locking a cross-process mutex for every request.
Fixed by 58dc4f0674066806c623a14dd47b1a5bc2ae3818

yias: seems reasonable 👌

yias (approved):
looks good, thanks.

PR #88, opened 2024-09-23
fix payload for surface_pp_input

yias:
## Description

This PR fixes the population of `surface_pp_input` when constructing the payload for the build request. `surface_pp_input` now gets constructed with the corresponding dictionaries of the `surface` variables.

story: [sc-24983]

## Test
- correcting the unit test
- smoke test against pre-prod


src/ansys/simai/core/data/model_configuration.py
```diff
@@ -440,7 +440,13 @@ def _to_payload(self):
...
             gcs = [asdict(gc) for gc in self.global_coefficients]
 
-        surface_pp_input_fld = self.pp_input.surface or []
+        surface_pp_input_fld = []
+        if self.pp_input.surface is not None:
+            surface_pp_input_fld = [
+                fd
+                for fd in sample_metadata.get("surface").get("fields")
+                if fd.get("name") in self.pp_input.surface
+            ]
```
awoimbee, line 449: Seems like we are silengthly ignoring invalid fields, but as the great @camzeech once said:

> silently ignoring a variable is not the right UX: we should inform the user either by warning or error.
By warning, there are consequences: the user can miss the warning, the build request (potentially useless) is actually sent and takes a license slot etc. I see no use case where a user wants willingly variables to be ignored: either it's a typo, or it's a fault (variable indeed does not exist in the data). In both cases, the user should review what they typed and input.
So I'd go with a proper error, preventing the build request in the first place.
Alternative suggestions always welcome :)

```suggestion
        surface_pp_input_fld = []
        if self.pp_input.surface is not None:
            sample_metadata_surface_fields = {v["name"]: v for v in sample_metadata["surface"]["fields"]}
            surface_pp_input_fld = [
                sample_metadata_surface_fields[name]
                for name in self.pp_input.surface
            ]
```

BTW `{}.get("toto").get("tata")` is worse than `{}["toto"]["tata"]`: one returns `AttributeError: 'NoneType' object has no attribute 'get'`, the other returns `KeyError: 'toto'`.
But I would be supportive of a function like:

```
def dict_get(obj: dict, *keys: str, default = None):
    for k in keys:
        obj = obj.get(k, {}) or {}
    return obj or default

```

yias: regarding the comment on "silently ignoring invalid fields"; yes this is something that it has been discussed already in sc-24271, and it is the topic of sc-24993, which would be an overall solution for all the keys (i.e., `surface`, `volume`, `surface_input`, `surface_pp_input`). This PR only fixes assigning `surface_pp_input` that is currently broken as sc-24983 describes.

yias: I updated the code implementing something like what you suggested in dict_get. Please have a look.

src/ansys/simai/core/utils/misc.py
```diff
@@ -34,3 +34,11 @@ def build_boundary_conditions(boundary_conditions: Optional[Dict[str, Any]] = No
     if not bc:
         raise ValueError("No boundary condition was specified.")
     return bc
+
+
+def dict_get(obj: dict, *keys: str, default=None):
+    """Get the requested key of the dictionary or opt to the default."""
+    for k in keys:
+        if k:
+            obj = obj.get(k, {}) or {}
```
awoimbee, line 43: why `if k` ?
```suggestion
        obj = obj.get(k, {}) or {}
```

src/ansys/simai/core/data/model_configuration.py
```diff
@@ -440,7 +441,13 @@ def _to_payload(self):
         if self.global_coefficients is not None:
             gcs = [asdict(gc) for gc in self.global_coefficients]
 
-        surface_pp_input_fld = self.pp_input.surface or []
+        surface_pp_input_fld = []
+        if self.pp_input.surface is not None:
+            sample_surface = dict_get(sample_metadata, "surface", {})
+            suface_fields = dict_get(sample_surface, "fields", default=[])
```
awoimbee, line 447:
```suggestion
            suface_fields = dict_get(sample_metadata, "surface", "fields", default=[])
```

src/ansys/simai/core/utils/misc.py
yias: oh, my bad; the assignment in `ModelConfiguration._to_payload` had a bug on the assignment. If clause is removed.

src/ansys/simai/core/data/model_configuration.py
yias: yup, and actually the previous assignment was packing `{}` too, that was creating problems. It's set now.

awoimbee (approved):
Let's go, time to merge

PR #99, opened 2024-10-09
Fix filters in `TrainingDataDirectory.list`

yias:
## Description

When setting a filter in  `TrainingDataDirectory.list`, the dictionaries in `to_raw_filters` are not encoded to a json object correctly. This causes the server to crash and the request to return a 500 server error. 

This PR fixes this issue with calling `json.dumps` to properly encode the disctionaries.

story: [sc-25499]

## Test
- tested against dev and pre-prod

### DIFF
--- a/src/ansys/simai/core/data/types.py
+++ b/src/ansys/simai/core/data/types.py
@@ -308,8 +308,7 @@ In both cases, the conditions are ``AND`` together.
 
 def to_raw_filters(filters: Optional["Filters"]) -> Optional["RawFilters"]:
     if isinstance(filters, dict):
-        tt = [{"field": k, "operator": "EQ", "value": v} for k, v in filters.items()]
-        return tt
+        return [{"field": k, "operator": "EQ", "value": v} for k, v in filters.items()]
     if isinstance(filters, list) and len(filters) > 0:
         if isinstance(filters[0], dict):
             return filters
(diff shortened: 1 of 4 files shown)


src/ansys/simai/core/data/types.py
```diff
@@ -308,9 +309,12 @@ class RawSingleFilter(TypedDict):
 
 def to_raw_filters(filters: Optional["Filters"]) -> Optional["RawFilters"]:
     if isinstance(filters, dict):
-        return [{"field": k, "operator": "EQ", "value": v} for k, v in filters.items()]
+        tt = [json.dumps({"field": k, "operator": "EQ", "value": v}) for k, v in filters.items()]
+        return tt
```
mmeisso, line 313: So it would return a `str[]` , thus breaking the return type of this function.

Function "to_raw_filters" is fine for me, it does it's job. In my opinion, it's in "iter_training_data" that you have to url_encode. ( maybe `query = urlencode([("filter[]", json.dumps(f)) for f in (filters or [])])` )

I'll go a bit further, but json.dumps produces an output that has spaces, so i'd rather add one parameter to dump :
```
json.dumps(data, separators=(',', ':'))
```

mmeisso (changes requested):

src/ansys/simai/core/data/types.py
yias: yup, I implemented it too fast. I moved the encoding to `TrainingDataClientMixin.iter_training_data`, where the gets prepared, added the separators.

mmeisso (approved):

PR #100, opened 2024-10-17
feat: Add support for non_parametric optimization

tmpbeing:


yias (approved):
It looks fine. I just have a few comments. 

I'm also wondering if a `run_parametric` and `run_non_parametric` could be a separated `dataclass`/`pydantic class` with fields and validators for avoiding having all the functions classified as internally used (`_`)

src/ansys/simai/core/data/optimizations.py
```diff
@@ -222,15 +217,122 @@ def my_geometry_generation_function(param_a, param_b):
...
+                    trial_run = optimization._run_iteration({})
+                    trial_run.wait()
+                    iteration_result = {
+                        "objective": trial_run.fields["outcome_values"],
+                    }
+                    progress_bar.set_postfix(**iteration_result)
+                    if trial_run.fields.get("is_feasible", True):
+                        iterations_results.append(iteration_result)
+                    else:
+                        logger.debug("Trial run results did not match constraints. Skipping.")
```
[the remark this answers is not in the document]
tmpbeing: Which is why it's in debug level. Most users won't ever see log from the SDK, you have to specifically enable them, even more so when it's debug.

src/ansys/simai/core/data/optimizations.py
```diff
@@ -222,15 +217,122 @@ def my_geometry_generation_function(param_a, param_b):
...
+                    progress_bar.update(1)
+                logger.debug("Optimization complete.")
+                progress_bar.set_description("Optimization complete.")
+            return iterations_results
+
 
-def _check_geometry_generation_fn_signature(geometry_generation_fn, geometry_parameters):
+def _validate_geometry_parameters(params: Dict):
+    if not isinstance(params, Dict):
+        raise InvalidArguments("geometry_parameters: must be a dict.")
```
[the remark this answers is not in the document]
tmpbeing: Sorry but I don't agree with those changes. In my mind they add a needless level of indirection. And while they would make changing the errors slightly easier, search/replace would do the job just fine.

src/ansys/simai/core/data/optimizations.py
```diff
@@ -222,15 +217,122 @@ def my_geometry_generation_function(param_a, param_b):
...
+                )
+            )
+            optimization.wait()
+            logger.debug("Optimization defined. Starting optimization loop.")
+            iterations_results: List[Dict] = []
+            with keep.running(on_fail="warn"):
+                for _ in range(n_iters):
+                    logger.debug("Running iteration")
+                    progress_bar.set_description("Running iteration")
+                    trial_run = optimization._run_iteration({})
```
kliment-slice, line 282: This `{}` arg could really use a validation with a raise `InvalidArguments`

src/ansys/simai/core/data/optimizations.py
```diff
@@ -239,16 +341,10 @@ def get(self, trial_run_id: str):
         """Get a specific trial run from the server."""
         return self._model_from(self._client._api.get_optimization_trial_run(trial_run_id))
 
-    def _try_geometry(
-        self,
-        optimization: Identifiable[Optimization],
-        geometry: Identifiable[Geometry],
-        geometry_parameters: Dict,
+    def _run_iteration(
```
kliment-slice, line 344: nit: same method name for both optim and optimtrialrun objects, but with different args..

kliment-slice (approved):

src/ansys/simai/core/data/optimizations.py
awoimbee, line 1: Since I could not wrap my head around the interaction of the different components I made this:
```mermaid
sequenceDiagram
    participant User
    participant OptDir as OptimizationDirectory
    participant TrialDir as OptimizationTrialRunDirectory

    User ->> OptDir: run_parametric
    activate User
    activate OptDir
    create participant Opti as Optimization
    OptDir -x Opti: creates


    loop Until optimization complete
        OptDir->>Opti: _run_iteration
        activate Opti
        Opti->>TrialDir: _run_iteration
        activate TrialDir
        create participant Trial as OptimizationTrialRun
        TrialDir-xTrial: Creates
        activate Trial
        TrialDir-->>Opti: returns OptimizationTrialRun
        deactivate TrialDir
        Opti-->>OptDir: returns OptimizationTrialRun
        deactivate Opti
        OptDir->>Trial: Get iteration results
        deactivate Trial
    end
    deactivate OptDir

    OptDir-->>User: result (List[Dict])
    deactivate User
```

Conclusions:
* Some objects are not meant to be handled by end users (OptimizationTrialRun and OptimizationTrialRunDirectory so should be prefixed by "_")
* The relationships between objects must be better defined in the code, even in the definition order, define the children first: `OptimizationTrialRun` then `OptimizationTrialRunDirectory` then `Optimization` then `OptimizationDirectory`
* Maybe `_run_iteration` should only be defined on the `Optimization` ? 
* I don't understand what `OptimizationTrialRunDirectory.get` could be used for

awoimbee (commented):
Generally nice but `data/optimizations.py` is hard to understand

PR #101, opened 2024-10-18
Cancel model build

kliment-slice:
[sc-25165]
Tested on dev1.

Caveats/Comments:
*while `/cancel-training` endpoint is called and logged (for the correct project, etc) and training is indeed cancelled, I do notice execution continues and a new workspace is still being created. 
`An error occurred during the prediction run: Dependency entered state failure` is also shown in the UI as a consequence of the canceled training. 
*Is this the desired behavior? Should we also be deleting the workspace (or `hidden=True`) with this `cancel_build()`?

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -29,6 +29,8 @@ jobs:
     name: "Documentation style"
     runs-on: ubuntu-latest
     steps:
+      - name: Install rst2html
+        run: sudo apt-get update && sudo apt-get install -y python-docutils
       - uses: ansys/actions/doc-style@v7
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -30,7 +30,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Install rst2html
-        run: sudo apt-get update && sudo apt-get install -y python-docutils
+        run: sudo apt-get update && sudo apt-get install -y python3-docutils
       - uses: ansys/actions/doc-style@v7
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -21,7 +21,7 @@ jobs:
     name: "Code style"
     runs-on: ubuntu-latest
     steps:
-      - uses: ansys/actions/code-style@v8
+      - uses: ansys/actions/code-style@v7
         with:
           python-version: ${{ env.MAIN_PYTHON_VERSION }}
 
@@ -43,7 +43,7 @@ jobs:
         os: [ubuntu-latest, windows-latest]
         python-version: ["3.9", "3.10", "3.11", "3.12"]
     steps:
-      - uses: ansys/actions/build-wheelhouse@v8
+      - uses: ansys/actions/build-wheelhouse@v7
         with:
           library-name: ${{ env.LIBRARY_NAME }}
           operating-system: ${{ matrix.os }}
@@ -66,7 +66,7 @@ jobs:
         run: pdm install --dev
       - name: Generate test requirements
         run: mkdir requirements && pdm export -G test --no-default -o requirements/requirements_tests.txt
-      - uses: ansys/actions/tests-pytest@v8
+      - uses: ansys/actions/tests-pytest@v7
         with:
           pytest-extra-args: "--cov=ansys --cov-report=term --cov-report=html:.cov/html"
           checkout: false
@@ -87,7 +87,7 @@ jobs:
       - uses: actions/checkout@v4
       - name: Generate doc requirements
         run: pdm export -G doc --no-default -o doc_requirements.txt
-      - uses: ansys/actions/doc-build@v8
+      - uses: ansys/actions/doc-build@v7
         with:
           python-version: ${{ env.MAIN_PYTHON_VERSION }}
           checkout: false
@@ -99,7 +99,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: [doc-build, tests]
     steps:
-      - uses: ansys/actions/build-library@v8
+      - uses: ansys/actions/build-library@v7
         with:
           library-name: ${{ env.LIBRARY_NAME }}
           python-version: ${{ env.MAIN_PYTHON_VERSION }}
@@ -110,7 +110,7 @@ jobs:
     needs: [build-library]
     if: github.event_name == 'push' && !contains(github.ref, 'refs/tags')
     steps:
-      - uses: ansys/actions/doc-deploy-dev@v8
+      - uses: ansys/actions/doc-deploy-dev@v7
         with:
           cname: ${{ env.DOCUMENTATION_CNAME }}
           token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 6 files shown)


yias (commented):
Good addition, thanks. I have some comments on the rationale. 

Regarding your question, I don't consider throwing a workspace error when cancelling a build as the expected behavior. Workspaces are generated right before the model's smoke test, which is late in the building process. Maybe, there is a mixup in the BE doesn't let the cancel request to be well processed. From a first glance, it doesn't look related to the SDK.

src/ansys/simai/core/data/models.py
```diff
@@ -108,3 +109,22 @@ def build(
...
+            model: a model object or model id whose ongoing build to cancel
+
+        .. code-block:: python
+
+            simai.models.cancel_build(new_model)
+        """
+
+        model_object = get_object_from_identifiable(model, self._client.models)
+        response = self._client._api.cancel_build(model_object.project_id)
+        return (
```
yias, line 126: what would the user do with the return strings? print them? Wouldn't it be more intuitive if just a boolean was returned?

yias (commented):
I'm not sure if a string or a boolean would be more intuitive for the user in `cancel_build`

yias (approved):
it looks fine

src/ansys/simai/core/data/models.py
```diff
@@ -108,3 +109,22 @@ def build(
                 dismiss_data_with_volume_overflow,
             )
         )
+
+    def cancel_build(self, model: Identifiable[Model]):
```
awoimbee, line 113: Why is this in ModelDirectory if the model is not used (the project is) ?
Imagine this:
```
old_model = ...
new_model_in_training = ...

client.models.cancel_training(old_model)  # oops I just cancelled the new one
```

kliment-slice: Agreed, moved to `projects`. ONLY reason for `models` was Camille's description of [sc-25165], where he points to models docs for the missing method.

.github/workflows/ci_cd.yml
```diff
@@ -29,6 +29,8 @@ jobs:
     name: "Documentation style"
     runs-on: ubuntu-latest
     steps:
+      - name: Install rst2html
```
awoimbee, line 32:
```suggestion
      # ansys/actions/doc-style@v7 is broken
      # TODO: update to v8 and remove apt-get commands
      - name: Install rst2html
```

src/ansys/simai/core/data/projects.py
```diff
@@ -255,3 +255,21 @@ def delete(self, project: Identifiable[Project]) -> None:
...
         self._client._api.delete_project(get_id_from_identifiable(project))
+
+    def cancel_build(self, project: Identifiable[Project]):
+        """Cancel an ongoing build if one exists.
+
+        Args:
+            project: ID or :class:`model <Project>` of the project.
+
+        Returns:
+            ``True`` if ``is_being_trained=False`` in the JSON response.
```
awoimbee, line 266: The end user doesn't care about "the JSON response".

Also in this case it seems like `is_being_trained=True` could never happen ? These cases are possible:
* project was training a model, we cancelled the training ->  `is_being_trained=False`
* project was not training a model -> `is_being_trained=False`
* an error happened -> exception is raised, `cancel_build` never returns

src/ansys/simai/core/data/projects.py
awoimbee, line 1: TODO: also add on `class Project` so one can:
```python
proj = client.projects.get(name="toto")
proj.cancel_training()

```

awoimbee (commented):
Thanks, much better, but I still have reservations

src/ansys/simai/core/data/projects.py
```diff
@@ -205,6 +205,14 @@ def compute_gc_formula(
 
         return gc_compute.result if gc_compute.is_ready else None
 
+    def cancel_training(self):
+        """Cancel training if there is one in progress."""
+
+        project_response = self._client._api.get_project(self.id)
```
awoimbee, line 211:
```suggestion
        self.reload()
```

But I think we should keep it simple and if we want the API to return if there actually was a training that got cancelled then we should do that instead of hacking around in the SDK.

src/ansys/simai/core/data/projects.py
```diff
@@ -205,6 +205,14 @@ def compute_gc_formula(
...
+        if project_response.get("is_being_trained") is False:
+            raise ProcessingError("No training in progress for this project.")
```
awoimbee, line 213: Why is it an error here but a no-op in ProjectDirectory ? Why is the return type different ?

tests/test_projects.py
```diff
@@ -307,3 +307,87 @@ def test_successful_gc_verify(simai_client):
...
+    assert failed_cancellation is False
+
+
+@responses.activate
+def test_cancel_active_training(simai_client):
+    """WHEN I call cancel_training() from a project with active training
+    THEN it succeeds
+    """
+
+    project = simai_client._project_directory._model_from({"id": "e45y123", "name": "proj"})
```
awoimbee, line 356:
```suggestion
    project = simai_client.projects._model_from({"id": "e45y123", "name": "proj"})
```

src/ansys/simai/core/data/projects.py
```diff
@@ -205,6 +205,16 @@ def compute_gc_formula(
 
         return gc_compute.result if gc_compute.is_ready else None
 
+    def cancel_training(self):
```
tmpbeing, line 208: Is it training ?

src/ansys/simai/core/data/projects.py
```diff
@@ -255,3 +265,26 @@ def delete(self, project: Identifiable[Project]) -> None:
             project: ID or :class:`model <Project>` of the project.
         """
         self._client._api.delete_project(get_id_from_identifiable(project))
+
+    def cancel_build(self, project: Identifiable[Project]):
```
tmpbeing, line 269: Or is it build ?
We need to pick. (To be transparent, I'm confused as well on which one we're supposed to use)

src/ansys/simai/core/data/projects.py
```diff
@@ -205,6 +205,16 @@ def compute_gc_formula(
 
         return gc_compute.result if gc_compute.is_ready else None
 
+    def cancel_training(self):
+        """Cancel training if there is one in progress."""
+
+        self.reload()
+        if self.fields.get("is_being_trained") is False:
+            return False
```
tmpbeing, line 213: Two things:
- I'm not sure we need to reload here, we could probably use `self.is_pending()`.
- Why do we return `False` ? This isn't documented in the types or the description and it doesn't make much sense. IMO this fucking should always return `None`. Arguably we could raise an error if the project/model isn't cancellable (if we don't already, not sure how the endpoint behaves).

So basically, I think this whole function could just be `self._client._api.cancel_build(self.id)`

src/ansys/simai/core/data/projects.py
```diff
@@ -255,3 +265,26 @@ def delete(self, project: Identifiable[Project]) -> None:
...
+            build_cancelled = simai.projects.cancel_build(project)
+        """
+
+        project_id = get_id_from_identifiable(project)
+        project_response = self._client._api.get_project(project_id)
+        if not project_response.get("is_being_trained"):
+            return False
+        cancel_response = self._client._api.cancel_build(project_id)
+
+        return not cancel_response.get("is_being_trained")
```
tmpbeing, line 290: Also could be simplified.

src/ansys/simai/core/data/projects.py
kliment-slice: - the endpoint is `projects/{project_id}/cancel-training`
- the UI refers to it as `Cancel Build` universally, ticket refers to it as `cancel build` and reference to `models.build` etc

src/ansys/simai/core/data/projects.py
kliment-slice: The endpoint responds 200 regardless of build status, so raising a `ProcessingError` makes sense.
Using `is_pending`: seems to return `True` for every project object..? (the model object would be the one pending during build). Thus, sticking to using `is_being_trained` field on the project; reload is indeed necessary (according to thorough live testing on dev1).

tmpbeing (approved):

awoimbee (approved):

PR #102, opened 2024-10-28
Fix examples

MorganeExtrality:
Some examples of the documentation present the following line of code:
```
simai = ansys.simai.core_from_config()
```

Which returns the following error:
```
Fix AttributeError: module 'ansys.simai' has no attribute 'core_from_config'
```

This PR fixes that.


tmpbeing (approved):
Thank you !

PR #104, opened 2024-11-04
Raise an error when a variable is not found in the reference sample

yias:
## Description

When an input/output/pp_input variable is introduced, the SDK now would validate that the variables exist in the reference sample. Otherwise, it would raise an error. The validation occurs upon assignment. 

This PR also does some small code improvements in `ModelConfiguration._to_payload`.

Story: sc-24993

### Example of the raised error
image

## Test
- Corresponding unit-test adding

### DIFF
--- a/src/ansys/simai/core/data/model_configuration.py
+++ b/src/ansys/simai/core/data/model_configuration.py
@@ -289,10 +289,7 @@ class ModelConfiguration:
 
     project: "Optional[Project]" = None
     continuous: bool = False
-    input: ModelInput = field(default_factory=lambda: ModelInput())
-    output: ModelOutput = field(default_factory=lambda: ModelOutput())
     domain_of_analysis: DomainOfAnalysis = field(default_factory=lambda: DomainOfAnalysis())
-    pp_input: PostProcessInput = field(default_factory=lambda: PostProcessInput())
 
     def __set_gc(self, gcs: list[GlobalCoefficientDefinition]):
         verified_gcs = []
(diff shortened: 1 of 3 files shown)


src/ansys/simai/core/data/model_configuration.py
```diff
@@ -330,6 +327,72 @@ def __get_build_preset(self):
...
+        sample_metadata = self.project.sample.fields.get("extracted_metadata")
+        if not sample_metadata.get("volume"):
+            raise ProcessingError(
+                "No volume file was found in the reference sample. A volume file is required to use volume variables."
+            ) from None
+        self.__validate_variables(vars_to_validate, "volume")
+
+    def __set_input(self, model_input: ModelInput):
+        if not model_input:
+            model_input = ModelInput()
```
tmpbeing, line 358: `model_input` is not marked as optional. Also with this if I do `model_config.input = None`, input is not gonna be `None` which is not what I would expect. Granted that `None` is not a valid value so maybe we should just raise instead ?

src/ansys/simai/core/data/model_configuration.py
```diff
@@ -330,6 +327,72 @@ def __get_build_preset(self):
...
+        self.__dict__["input"] = model_input
+
+    def __get_input(self):
+        return self.__dict__.get("input")
+
+    input = property(__get_input, __set_input)
+
+    def __set_output(self, model_output: ModelOutput):
+        if not model_output:
+            model_output = ModelOutput()
```
tmpbeing, line 370: Likewise

src/ansys/simai/core/data/model_configuration.py
yias: `input` is set as optional in the constructor of `ModelConfiguration` see [here](https://github.com/ansys/pysimai/blob/9d80677d6e297bade43d4f30fe8dc7d35388b262/src/ansys/simai/core/data/model_configuration.py#L342), so it could get an empty `ModelInput` but not `None` indeed. I changed it so it raises an error instead

src/ansys/simai/core/data/model_configuration.py
yias: changed too, to raise an error.

src/ansys/simai/core/data/model_configuration.py
```diff
@@ -330,6 +327,72 @@ def __get_build_preset(self):
...
+        if unknown_variables:
+            raise ProcessingError(
+                f"{self.__class__.__name__}: {var_type} variables {', '.join(unknown_variables)} do not exist in the reference sample."
+            ) from None
+
+    def __validate_surface_variables(self, vars_to_validate: list[str]):
+        sample_metadata = self.project.sample.fields.get("extracted_metadata")
+        if not sample_metadata.get("volume"):
+            raise ProcessingError(
+                "No volume file was found in the reference sample. A volume file is required to use volume variables."
```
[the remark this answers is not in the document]
yias: oops. code is updated for surface variables.

src/ansys/simai/core/data/model_configuration.py
```diff
@@ -330,6 +327,76 @@ def __get_build_preset(self):
...
 
+    def __validate_variables(self, vars_to_validate: list[str], var_type: str):
+        sample_metadata = self.project.sample.fields.get("extracted_metadata")
+        var_fields = dict_get(sample_metadata, var_type, "fields", default=[])
+        var_fields_name = {fd.get("name") for fd in var_fields}
+        unknown_variables = set(vars_to_validate) - var_fields_name
+        if unknown_variables:
+            raise ProcessingError(
+                f"{self.__class__.__name__}: {var_type} variables {', '.join(unknown_variables)} do not exist in the reference sample."
+            ) from None
```
tmpbeing, line 338: AFAIK `from None` is only useful in a context where an exception could already have been thrown (in the `except` of a `try`)
https://docs.python.org/3/tutorial/errors.html#exception-chaining
I don't think this is the care here.

src/ansys/simai/core/data/model_configuration.py
```diff
@@ -330,6 +327,76 @@ def __get_build_preset(self):
...
+
+    def __validate_surface_variables(self, vars_to_validate: list[str]):
+        sample_metadata = self.project.sample.fields.get("extracted_metadata")
+        if not sample_metadata.get("surface"):
+            raise ProcessingError(
+                "No surface field is found in the reference sample. A surface field is required to use surface variables."
+            ) from None
```
tmpbeing, line 345: Same

src/ansys/simai/core/data/model_configuration.py
```diff
@@ -330,6 +327,76 @@ def __get_build_preset(self):
...
+        self.__validate_variables(vars_to_validate, "surface")
+
+    def __validate_volume_variables(self, vars_to_validate: list[str]):
+        sample_metadata = self.project.sample.fields.get("extracted_metadata")
+        if not sample_metadata.get("volume"):
+            raise ProcessingError(
+                "No volume field is found in the reference sample. A volume field is required to use volume variables."
+            ) from None
```
tmpbeing, line 353: Same

tmpbeing (approved):
LGTM

src/ansys/simai/core/data/model_configuration.py
yias: exception chaining removed

PR #106, opened 2024-11-05
Fix wrong type hints on SimAIClient

### LINKED ISSUE #103: Erroneous type hints for `SimAIClient`
### 🔍 Before submitting the issue

- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### 🐞 Description of the bug

When instantiating a `SimAIClient`, the type checker (Pyright) displays the initializer's parameter types incorrectly. This can be reproduced by running Pyright from the command line, as well as the type hints that VS Code displays on hovering over the instance declaration.

### 📝 Steps to reproduce

Instantiate a SimAI client as follows:

```py
from ansys.simai.core.client import SimAIClient


_ = SimAIClient(
    organization="foo",
    credentials={
        "username": "bar",
        "password": "baz",
    },
)
```

Run Pyright in strict mode and observe the following errors:

```
/workspaces/test.py
  /workspaces/test.py:5:18 - error: Argument of type "Literal['foo']" cannot be assigned to parameter "organization" of type "Credentials | None"
    Type "Literal['foo']" is not assignable to type "Credentials | None"
      "Literal['foo']" is not assignable to "Credentials"
      "Literal['foo']" is not assignable to "None" (reportArgumentType)
  /workspaces/test.py:6:17 - error: Argument of type "dict[str, str]" cannot be assigned to parameter "credentials" of type "str | None"
    Type "dict[str, str]" is not assignable to type "str | None"
      "dict[str, str]" is not assignable to "str"
      "dict[str, str]" is not assignable to "None" (reportArgumentType)
```

When hovering over the `SimAIClient` type, VS Code is also displaying erroneous type hints:

Image

### 💻 Which operating system are you using?

Linux

### 📀 Which ANSYS version are you using?

No ANSYS version needed, the error occurs within the SDK

### 🐍 Which Python version are you using?

3.12

### 📦 Installed packages

ansys-simai-core==0.2.4

awoimbee:
Fixes #103

The issue is that in:
```python
    @steal_kwargs_type(ClientConfig)
    def __init__(self, **kwargs):
```
The `self` arg is causing an issue as it is not present in `ClientConfig`.

My fix: `Concatenate[Any, T]`: type `self` as `Any`.
We would need python >=3.12 to properly type `self`, using something like `@steal_kwargs_type_on_method[SimAIClient, ClientConfig]()`

### DIFF
--- a/src/ansys/simai/core/utils/typing.py
+++ b/src/ansys/simai/core/utils/typing.py
@@ -20,9 +20,11 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-from typing import TYPE_CHECKING, Any, Callable, Concatenate, ParamSpec, TypeVar, cast
+from typing import TYPE_CHECKING, Any, Callable, ParamSpec, TypeVar, cast
 
 if TYPE_CHECKING:
+    from typing_extensions import Concatenate
+
     P = ParamSpec("P")
     T = TypeVar("T")
 
--- a/src/ansys/simai/core/utils/typing.py
+++ b/src/ansys/simai/core/utils/typing.py
@@ -20,10 +20,10 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-from typing import TYPE_CHECKING, Any, Callable, ParamSpec, TypeVar, cast
+from typing import TYPE_CHECKING, Any, Callable, TypeVar, cast
 
 if TYPE_CHECKING:
-    from typing_extensions import Concatenate
+    from typing_extensions import Concatenate, ParamSpec
 
     P = ParamSpec("P")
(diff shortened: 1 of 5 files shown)


src/ansys/simai/core/utils/typing.py
```diff
@@ -20,21 +20,21 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-from typing import TYPE_CHECKING, cast
+import typing as t
```
tmpbeing, line 23: Convenient but all the `t.{type}`s makes the thing less readable IMO

tmpbeing (approved):
Ah yes makes sense. Thought it was maybe an invisible`*args` causing the issue but `self` makes more sense.

src/ansys/simai/core/utils/typing.py
yias: I agree on this. I'd prefer either to import it without an alias or import just the required functions (as we generally do in this repo).

yias (approved):

PR #107, opened 2024-11-07
Fix example

MorganeExtrality:
Hello!
I noticed in the example that the constraint and the objective were the same global coefficient, but it is not possible. 
Here is a fix. 

Thanks


tmpbeing (approved):
Thanks

PR #110, opened 2024-11-19
Check new version from pypi instead of our api: fix pysimai version check

awoimbee:
:ticket:  https://app.shortcut.com/simai/story/26689/pysimai-fetch-latest-release-from-pypi

Our API stopped containing the latest pysimai version when it moved to this repo.
Since it's now on pypi we can just query there instead of querying our API.

I kept `warn_package_outdated` quite generic so internal packages can use it too (just like they did with `_check_for_new_version` before).

### DIFF
--- a/src/ansys/simai/core/client.py
+++ b/src/ansys/simai/core/client.py
@@ -24,7 +24,6 @@ import logging
 from typing import List, Optional
 
 from pydantic import ValidationError
-from semver.version import Version
 
 from ansys.simai.core import __version__
 from ansys.simai.core.api.client import ApiClient
@@ -54,6 +53,7 @@ from ansys.simai.core.errors import (
 )
 from ansys.simai.core.utils.config_file import get_config
 from ansys.simai.core.utils.configuration import ClientConfig
+from ansys.simai.core.utils.misc import warn_package_outdated
 from ansys.simai.core.utils.typing import steal_kwargs_type_on_method
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


src/ansys/simai/core/utils/misc.py
```diff
@@ -41,3 +48,24 @@ def dict_get(obj: dict, *keys: str, default=None):
...
+        logger.debug(f"Could not parse package version: {e}")
+    else:
+        if version_current < version_latest:
+            warn_template = (
+                f"A new version of {package} is %s. "
+                "Upgrade to get new features and ensure compatibility with the server."
+            )
+            if (
+                version_current.major < version_latest.major
+                or version_current.minor < version_latest.minor
```
tmpbeing, line 67: I still don't like that we're raising an error on minor change. I guess it kind of works for now cause we're on 0.x.x so i'll just not bump the minor unless there's big breaking changes. But randomly breaking all the scripts is bad IMO.

src/ansys/simai/core/client.py
```diff
@@ -324,26 +324,15 @@ def wait(self) -> None:
                 raise errors[0]
             raise MultipleErrors(errors)
 
-    def _check_for_new_version(self, client_name="ansys.simai.core", current_version=__version__):
+    def _check_for_new_version(self) -> None:
```
tmpbeing, line 327: FYI this will break napoleon.

tmpbeing (approved):

src/ansys/simai/core/utils/misc.py
awoimbee: Agree, cahnged to a "critical" log

PR #112, opened 2024-11-26
Add options for using custom TLS CA bundles

awoimbee:
https://app.shortcut.com/simai/story/26712/tls-certs-option-to-use-system-certs-to-disable-validation

Allows users to:
* use system CA certs (should fix issues with <https://zscaler.com>)
* provide a custom CA bundle without using `REQUESTS_CA_BUNDLE`
* disable verification (for troubleshooting)

### DIFF
--- a/src/ansys/simai/core/api/mixin.py
+++ b/src/ansys/simai/core/api/mixin.py
@@ -62,13 +62,12 @@ class ApiClientMixin:
 
     def __init__(self, *args, config: ClientConfig):  # noqa: D107
         self._session = requests.Session()
-        match config.tls_ca_bundle:
-            case "system":
-                self._session.mount("https://", TruststoreAdapter())
-            case "unsecure-none":
-                self._session.verify = False
-            case os.PathLike():
-                self._session.verify = str(config.tls_ca_bundle)
+        if config.tls_ca_bundle == "system":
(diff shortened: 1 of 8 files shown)


tests/test_client_config_tls_ca_bundle.py
```diff
@@ -0,0 +1,144 @@
+# Copyright (C) 2023 ANSYS, Inc. and/or its affiliates.
```
tmpbeing, line 1:
```suggestion
# Copyright (C) 2024 ANSYS, Inc. and/or its affiliates.
```
Though I don't usually bother with tests as they're not distributed.

src/ansys/simai/core/api/mixin.py
```diff
@@ -43,11 +46,28 @@
 logger = logging.getLogger(__name__)
 
 
+class TruststoreAdapter(HTTPAdapter):
+    def init_poolmanager(self, *a, **kw):
+        if sys.version_info < (3, 10):
+            raise ConfigurationError("The system CA store can only be used with python >= 3.10")
```
tmpbeing, line 52: Might be worth adding that limitation to the documentation

tmpbeing (approved):
Nice

yias (approved):
looks fine

PR #116, opened 2024-12-12
Remove version property from model manifest

awoimbee:
https://app.shortcut.com/simai/story/27197/remove-version-from-workspace-model-manifest


yias (approved):
Better like that, thanks!

PR #117, opened 2024-12-24
[feat] Support post processing predict as learnt and predict on cells for surface variables

yias:
## Description

This PR enables users to trigger post-processing on surface variables letting the original data association of the sample; e.g., data associated with points will continue be associated with points. 

To enable this option, the class `SurfaceVTPTDLocation` is introduced, together with the methods `PredictionPostProcessings.surface_vtp_td_location` and `SelectionPostProcessingsMethods.surface_vtp_td_location`. These methods trigger a post-processing on the surface variables while keeping the original data association.

Story: [sc-26496]

## Tests
- Unit-tests added
- Interactive smoke testing against dev env.

### DIFF
--- a/doc/source/api_reference/post_processings.rst
+++ b/doc/source/api_reference/post_processings.rst
@@ -56,6 +56,14 @@ Available postprocessings
     :members:
 
 
+.. autoclass:: PostProcessingOnCells()
+    :members:
+
+
+.. autoclass:: PostProcessingAsLearnt()
+    :members:
+
+
 .. autoclass:: VolumeVTU()
     :members:
 
@@ -68,4 +76,7 @@ Helpers
 -------
 
 .. autoclass:: DownloadableResult()
+    :members:
+
+.. autoclass:: PPSurfaceLocation()
     :members:
--- a/doc/source/api_reference/post_processings.rst
+++ b/doc/source/api_reference/post_processings.rst
@@ -52,11 +52,11 @@ Available postprocessings
     :members:
 
 
-.. autoclass:: PostProcessingOnCells()
+.. autoclass:: SurfaceVTP()
     :members:
 
 
-.. autoclass:: PostProcessingAsLearnt()
+.. autoclass:: SurfaceVTPTDLocation()
     :members:
 
 
@@ -72,7 +72,4 @@ Helpers
 -------
 
 .. autoclass:: DownloadableResult()
-    :members:
-
-.. autoclass:: PPSurfaceLocation()
     :members:
--- a/doc/source/api_reference/post_processings.rst
+++ b/doc/source/api_reference/post_processings.rst
@@ -52,10 +52,6 @@ Available postprocessings
     :members:
 
 
-.. autoclass:: SurfaceVTP()
-    :members:
-
-
 .. autoclass:: PostProcessingOnCells()
     :members:
 
--- a/src/ansys/simai/core/data/post_processings.py
+++ b/src/ansys/simai/core/data/post_processings.py
@@ -25,7 +25,6 @@ import logging
 import numbers
 import sys
 from abc import ABC, abstractmethod
-from enum import Enum
 from inspect import cleandoc
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Type, Union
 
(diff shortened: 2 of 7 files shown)


doc/source/api_reference/post_processings.rst
```diff
@@ -56,6 +56,14 @@ Available postprocessings
     :members:
 
 
+.. autoclass:: PostProcessingOnCells()
+    :members:
```
tmpbeing, line 60: Might want to have inherited-members as well.

src/ansys/simai/core/data/post_processings.py
```diff
@@ -274,6 +275,40 @@ class SurfaceVTP(_PostProcessingVTKExport):
     """
 
 
+class PostProcessingOnCells(SurfaceVTP):
```
tmpbeing, line 278: I didn't follow the product story but these class name seems too generic. Maybe `SurfaceOnCells` and `SurfaceAsLearnt` ?

tmpbeing (commented):
I think class names need to change. And we might want to turn `SurfaceVTP` into `_SurfaceVTP` and undocument it if it's not meant to be used directly anymore. Hell maybe remove it entirely given the class is empty.

src/ansys/simai/core/data/post_processings.py
yias: My intention is to follow the naming as in the description of the story; since the post-processing is described as `predict on cells`, where `cells` are the data under the `cell` keyword in pyvista objects, I assume that this name would be self-explanatory to users. Nevertheless, please feel free to bring any suggestions on.

awoimbee (commented):
> Conceptually, both objects of PostProcessingOnCells and of PostProcessingAsLearnt() are SurfaceVTP objects, since the resource is a surface vtp file. Their only deference is the endpoint. If we need to do any processing on the returned (by both options) surface vtp file, it would be realized in the SurfaceVTP class.

But the `SurfaceVTP` class is an empty shell here ? So it does not do any processing ?

src/ansys/simai/core/data/post_processings.py
```diff
@@ -425,7 +460,9 @@ def slice(
         plane = convert_axis_and_coordinate_to_plane_eq_coeffs(axis, coordinate)
         return self._get_or_run(Slice, {"plane": plane, "output_format": format}, run)
 
-    def surface_vtp(self, run: bool = True) -> Optional[SurfaceVTP]:
+    def surface_vtp(
+        self, run: bool = True, pp_location: PPSurfaceLocation = PPSurfaceLocation.ON_CELLS
```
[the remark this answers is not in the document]
yias: An enum might require a little bit of more coding from the part of the users, but it is a more structured way to control the input of arguments, than strings. This is way I used it here.

src/ansys/simai/core/data/post_processings.py
yias: PM has accepted the suggestion on the classes names, resolving the comment.

yias:
class `SurfaceVTP` has been renamed to `_SurfaceVTP` and removed from documentation. Suggestion on class names is accepted by PM. Please review.

tmpbeing (approved):
If PM is happy

yias:
I just realized that `SurfaceVTP` is the type of post-processing that the BE returns in the response body when requesting surface post-processing. Since the method `.model_from` uses the type of the response for creating instances of this type, `SurfaceVTP` is tight to the BE response. To me, this qualifies it as a class name, so I'm bringing back its original name (by removing `_`).

tmpbeing (changes requested):
> I just realized that SurfaceVTP is the type of post-processing that the BE returns in the response body when requesting surface post-processing. Since the method .model_from uses the type of the response for creating instances of this type, SurfaceVTP is tight to the BE response. To me, this qualifies it as a class name, so I'm bringing back its original name (by removing _).

Ok this is actually a bit of a red flag. If I understand correctly that means we create different classes in different contexts for the same thing. Looking at `model_from` and specifically `_data_model_for_type_name` it seems we'll get an error if we get a `SurfaceVTPTDLocation` from the backend.

src/ansys/simai/core/data/post_processings.py
misterjoa: Hello everyone! I was just reading this PR when I fell upon this thread. It seems that comments have been discarded without finding agreement between everyone. Let's discuss this together this afternoon, I'll schedule a meeting!

yias:
@tmpbeing We would get an error if BE would return `SurfaceVTPTDLocation`, however it does not return `SurfaceVTPTDLocation`. It returns `SurfaceVTP` for post-processing on surface vars (both for predict as learnt and predict on cells), and this is why it doesn't break. Let me know what you think

src/ansys/simai/core/data/post_processings.py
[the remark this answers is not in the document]
yias: since a new method is introduced, this discussion is outdated. Thus, I'm resolving it

yias:
PR is updated as discussed in our meeting. Reviews are welcomed, thanks.

awoimbee (approved):

PR #124, opened 2025-02-25
add clt.training_data.iter()

awoimbee:
Because waiting for potentially tens of thousands of objects to be fetched into a list isn't great UX.
With this change one can still to `len(clt.training_data.iter())`, it will just take 1/100th of the time to do so.


tmpbeing (approved):

tmpbeing:
Rebase on master to fix the documentation build CI

PR #135, opened 2025-04-04
Actually retry on HTTP 5xx

awoimbee:


src/ansys/simai/core/api/mixin.py
```diff
@@ -147,6 +149,17 @@ def _request(
...
+            m = re.search("too many ([0-9]{3}) error responses", str(e))
+            if m:
+                code = m.group(1)
+                if code == "502":
+                    raise ApiClientError("502 Bad Gateway") from None
+                if code == "503":
+                    raise ApiClientError("503 Service Unavailable") from None
+                if code == "504":
+                    raise ApiClientError("504 Gateway Timeout") from None
+            raise ApiClientError(str(e)) from None
```
awoimbee, line 162: That's obviously freaking terrible but otherwise the error looks like:
```
None: Max retries exceeded with url: https://test.test/502 (Caused by ResponseError('too many 502 error responses'))
```
And friday evening is not the time for exceptions redesign :)
WDYT ?

tmpbeing (approved):
Thanks ! Nice improvement.

PR #139, opened 2025-04-16
feat: add checks on "build on top"

msd-11:
Currently, no check is done when building a model with continuous learning feature.
This PR add checks for model configuration changes and for the continuous learning capabilities returned from API.

### DIFF
--- a/src/ansys/simai/core/api/project.py
+++ b/src/ansys/simai/core/api/project.py
@@ -95,7 +95,7 @@ class ProjectClientMixin(ApiClientMixin):
         dismiss_data_with_fields_discrepancies: bool = False,
         dismiss_data_with_volume_overflow: bool = False,
     ):
-        """Launches a build on top of previous model for a project.
+        """Launches a build on top of the previous model for a project.
 
         Args:
             project_id: the ID of the project
--- a/src/ansys/simai/core/api/project.py
+++ b/src/ansys/simai/core/api/project.py
@@ -89,6 +89,26 @@ class ProjectClientMixin(ApiClientMixin):
         }
         return self._post(f"projects/{project_id}/model", json=config, params=params)
 
+    def launch_build_on_top(
+        self,
+        project_id: str,
+        dismiss_data_with_fields_discrepancies: bool = False,
+        dismiss_data_with_volume_overflow: bool = False,
+    ):
+        """Launches a build on top of previous model for a project.
+
+        Args:
+            project_id: the ID of the project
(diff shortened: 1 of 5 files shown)


src/ansys/simai/core/data/models.py
```diff
@@ -107,6 +107,23 @@ def build(
         if not is_trainable:
             raise InvalidArguments(f"Cannot train model because: {is_trainable.reason}")
 
+        # Check if there is a difference between the two model configuration (except continuous field)
+        configuration_mismatch = not configuration.project.last_model_configuration or {
+            k
+            for k, _ in configuration.project.last_model_configuration._to_payload().items()
+            ^ configuration._to_payload().items()
+        } != {"continuous"}
```
mmeisso, line 115: I think we should just open new endpoint, one POST /model/on-top , that would take 0 arguments, and the current one would NOT accept build on top. Exit this comparison.

src/ansys/simai/core/api/project.py
```diff
@@ -89,6 +89,26 @@ def launch_build(
         }
         return self._post(f"projects/{project_id}/model", json=config, params=params)
 
+    def launch_build_on_top(
+        self,
+        project_id: str,
+        dismiss_data_with_fields_discrepancies: bool = False,
+        dismiss_data_with_volume_overflow: bool = False,
+    ):
+        """Launches a build on top of previous model for a project.
```
awoimbee, line 98:
```suggestion
        """Launches a build on top of the previous model for a project.
```

src/ansys/simai/core/data/models.py
```diff
@@ -107,11 +107,20 @@ def build(
...
+            )
+        else:
+            return self._model_from(
+                self._client._api.launch_build(
+                    configuration.project.id,
+                    configuration._to_payload(),
+                    dismiss_data_with_fields_discrepancies,
+                    dismiss_data_with_volume_overflow,
+                )
             )
```
awoimbee, line 126: nit: style
```suggestion
        if configuration.build_on_top:
            response = self._client._api.launch_build_on_top(
                configuration.project.id,
                dismiss_data_with_fields_discrepancies,
                dismiss_data_with_volume_overflow,
            )
        else:
            response = self._client._api.launch_build(
                configuration.project.id,
                configuration._to_payload(),
                dismiss_data_with_fields_discrepancies,
                dismiss_data_with_volume_overflow,
            )
        return self._model_from(response)
```

awoimbee (approved):

PR #141, opened 2025-04-28
feat: user api wont return record for jobs with pending status

u8slvn:
In upcoming versions, the user_api will no longer return record data updates for jobs in pending states, as these states don't contain meaningful changes. Additionally, job state information will be retrieved directly from the `status` field.


tmpbeing (approved):
Thanks !

mmeisso (approved):

PR #149, opened 2025-05-21
feat: implement process formula endpoint

u8slvn:
This PR aims to implement the new `process-formula` endpoint that offers a single route to check, compute, and cache formulas.

In order to prevent further modification of the codebase and accommodate eventual future GC features, I left the `GlobalCoefficientRequestDirectory` intact, even though there is only one child class now.

I also introduced a `delayed_events` pytest fixture that helps with testing SSE event. It can be used to schedule SSE events before calling the `wait` method without requiring heavy mocking.

---

:warning: This feature introduces breaking changes as `verify_gc_formula` and `compute_gc_formula` methods of `Project` DataModel class are replaced by `process_gc_formula`.

:no_entry: DO NOT MERGE until the endpoint has been released on User API.

### DIFF
--- a/src/ansys/simai/core/data/global_coefficients_requests.py
+++ b/src/ansys/simai/core/data/global_coefficients_requests.py
@@ -22,7 +22,7 @@
 
 import logging
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING, Any, TypeVar, Union
+from typing import TYPE_CHECKING, Any, Optional, TypeVar, Union
 
 from ansys.simai.core.data.base import (
     ERROR_STATES,
@@ -63,7 +63,7 @@ class GlobalCoefficientRequest(ABC, ComputableDataModel):
         )
 
     @ComputableDataModel._failure_message.getter
-    def _failure_message(self) -> str:
+    def _failure_message(self) -> Optional[str]:
         return self.fields.get("error")
 
     def _compose_calculette_payload(
--- a/src/ansys/simai/core/utils/requests.py
+++ b/src/ansys/simai/core/utils/requests.py
@@ -23,7 +23,7 @@
 import logging
 from http import HTTPStatus
 from json.decoder import JSONDecodeError
-from typing import Literal, Union, overload
+from typing import Literal, overload
 
 import niquests
 
@@ -82,9 +82,7 @@ def handle_http_errors(response: niquests.Response) -> None:
 
 
 @overload
-def handle_response(
-    response: niquests.Response, return_json: Literal[True]
-) -> Union[JSON, None]: ...
+def handle_response(response: niquests.Response, return_json: Literal[True]) -> JSON: ...
 
 
 @overload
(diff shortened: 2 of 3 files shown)


src/ansys/simai/core/utils/requests.py
```diff
@@ -108,13 +109,13 @@ def handle_response(response: requests.Response, return_json: bool = True) -> AP
     handle_http_errors(response)
 
     logger.debug("Returning response.")
-    if return_json:
+    if return_json and response.status_code != HTTPStatus.NO_CONTENT:
```
awoimbee, line 112: Seems like we should return `None` instead of `response` ?
Plz check if the `@overload` with `return_json: True` is still correct

tests/conftest.py
```diff
@@ -234,3 +232,65 @@ def _factory(**kwargs) -> Model:
...
+            # Make assertions
+            assert result is True
+    """
+
+    class DelayedEventManager:
+        def __init__(self) -> None:
+            self.events = []
+            self.thread = None
+
+        def add(self, delay: float, callback: Callable[[], None] = None):
```
awoimbee, line 266: IMO make delay optional with a 0.1 default. The important part about this fixture is that the events happen in order (otherwise we would just use <https://docs.python.org/2/library/threading.html#timer-objects>).

-> it should be more obvious in the doc that the operations are in order and `delay` is relative.

awoimbee (commented):
I have comments, but thanks for the change !

src/ansys/simai/core/data/global_coefficients_requests.py
```diff
@@ -112,9 +112,11 @@ def get(self, item_id: str) -> GlobalCoefficientRequestType:
 
         return self._registry[item_id] if item_id not in self._registry else None
 
-    def _handle_sse_event(self, data):
-        gc_formula = data.get("target").get("formula")
-        action = data.get("target").get("action")
+    def _handle_sse_event(self, data: dict[str, Any]) -> None:
+        gc_formula = data.get("target", {}).get("formula")
+        action = data.get("target", {}).get("action")
+        # Set the action to process if it is check or compute.
```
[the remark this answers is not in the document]
u8slvn: Action definition is set up in the SSE part. But now `check` and `compute` are manage within the same directory, which is `ProcessGlobalCoefficientDirectory`. So it's now only one item in the registry, so one ID, it's why `check` and `compute` are transformed into `process`. 

I will write a better comment.

src/ansys/simai/core/utils/requests.py
u8slvn: I agree, I'll check if there are no other side effects.

tests/conftest.py
u8slvn: Good idea, I'll make the update.

src/ansys/simai/core/utils/requests.py
```diff
@@ -94,7 +97,9 @@ def handle_response(
 def handle_response(response: niquests.Response, return_json: bool) -> APIResponse: ...
 
 
-def handle_response(response: niquests.Response, return_json: bool = True) -> APIResponse:
+def handle_response(
+    response: niquests.Response, return_json: bool = True
+) -> Union[APIResponse, None]:
```
awoimbee, line 102: That's not good, the typing of _get, _patch, ... is now false.
I think the right move is to do: `JSON = Union[Dict[str, Any], List[Dict[str, Any]], None]`

awoimbee (approved):

src/ansys/simai/core/utils/requests.py
u8slvn: Well spotted, I didn't see the `JSON` type declaration, thanks!
Update done.

PR #151, opened 2025-05-26
docs: update code examples for the optimization parametric and non parametric

msd-11:
Optimization can be run using both parametric and non-parametric methods. This updates the documentation examples for both approaches.

### DIFF
--- a/src/ansys/simai/core/data/optimizations.py
+++ b/src/ansys/simai/core/data/optimizations.py
@@ -276,9 +276,10 @@ class OptimizationDirectory(Directory[Optimization]):
             import ansys.simai.core
 
             simai = ansys.simai.core.from_config(workspace="optimization-workspace")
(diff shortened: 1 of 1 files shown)


src/ansys/simai/core/data/optimizations.py
```diff
@@ -269,6 +269,23 @@ def run_non_parametric(
...
+
+        Example:
+          .. code-block:: python
+
+            import ansys.simai.core
+
+            simai = ansys.simai.core.from_config(workspace="optimization-workspace")
+
+            simai.optimizations.run_non_parametric(
+                geometry="/local/path/to/geometry.vtp",
```
tmpbeing, line 281: Not a path, it's an `identifiable[geometry]` (an id or geometry object)

msd-11: Good catch ! Thx

PR #152, opened 2025-06-02
feat: improve check for parametric and non parametric optimizations

u8slvn:
Add various validations to enforce checks for running parametric and non-parametric optimizations.

### DIFF
--- a/src/ansys/simai/core/data/optimizations.py
+++ b/src/ansys/simai/core/data/optimizations.py
@@ -176,15 +176,19 @@ class OptimizationDirectory(Directory[Optimization]):
             )
         """
         workspace_id = get_id_from_identifiable(workspace, True, self._client._current_workspace)
+        outcome_constraints = outcome_constraints or []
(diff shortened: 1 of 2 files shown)


tmpbeing (approved):
Thanks.

src/ansys/simai/core/data/optimizations.py
```diff
@@ -176,15 +176,19 @@ def my_geometry_generation_function(param_a, param_b):
             )
         """
         workspace_id = get_id_from_identifiable(workspace, True, self._client._current_workspace)
+        outcome_constraints = outcome_constraints or []
+        workspace = workspace or self._client._current_workspace
```
msd-11, line 180:
```suggestion
```
This line seems unnecessary

msd-11 (approved):
LGTM

src/ansys/simai/core/data/optimizations.py
u8slvn: Nice catch! I'll remove it!

PR #155, opened 2025-06-04
feat: geomai preview

tmpbeing:
Note: Missing some tests currently. Will add them as I have time but I want the code reviewed so we can start pushing it.


src/ansys/simai/core/data/geomai/models.py
```diff
@@ -0,0 +1,165 @@
...
+    The preset to use for the model training duration. One of `debug`, `short`, `default`, `long`.
+
+    Mutually exclusive with `nb_epochs`.
+    """
+    nb_epochs: Optional[int] = Field(default=None, ge=1, le=300)
+    """
+    The number of epochs the training will run for, between 1 and 300.
+
+    Mutually exclusive with `build_preset`.
+    """
```
tmpbeing, line 49: @marielelandais 
I need your help for documenting these.

doc/source/api_reference/geomai/training_data.rst
```diff
@@ -0,0 +1,23 @@
+.. _geomai_training_data:
+
+GeomAITrainingData
+==================
+
+.. py:module:: ansys.simai.core.data.geomai.training_data
+
+A :class:`GeomAITrainingData<ansys.simai.core.data.geomai.training_data.GeomAITrainingData>` instance is a
+collection of :class:`GeomAITrainingDataPart<ansys.simai.core.data.geomai.training_data_parts.GeomAITrainingDataPart>`
+instances representing a prediction that can be used as input for the training of models.
```
awoimbee, line 10:
```
. In most cases it will contain a single part: the geometry to train on.
```

src/ansys/simai/core/api/geomai/predictions.py
awoimbee, line 1: I can't wait for all of this to be generated from the openapi schema

awoimbee (approved):
LGTM

src/ansys/simai/core/data/geomai/models.py
```diff
@@ -0,0 +1,160 @@
...
+from ansys.simai.core.data.types import Identifiable, get_id_from_identifiable
+from ansys.simai.core.errors import InvalidArguments
+
+if TYPE_CHECKING:
+    from ansys.simai.core.data.geomai.projects import GeomAIProject
+
+
+class GeomAIModelConfiguration(BaseModel):
+    nb_latent_param: int = Field(ge=2, le=256)
+    "TODO document"
```
marielelandais, line 37: """
    This number defines the number of floats that will be listed in the `latent_params` parameter for prediction. 
    It has to be defined according to the complexity and the diversity of the geometries you used as training data.

    You need to find the smallest number of latent parameters enabling the model to rebuild 
    a training data with a similar level of detail as the original geometry.

    - If the number is too low, the generated geometries will be too coarse.
    - If the number is too high, the model will not be able to generate consistent geometries.

    In most cases, the optimal number of latent parameters is lower than `20`.
    """
`+`
we could add a link to `GeomAIPredictionConfiguration` > `latent_params` at the end of the first sentence?

src/ansys/simai/core/data/geomai/models.py
```diff
@@ -0,0 +1,160 @@
...
+    build_preset: Optional[Literal["debug", "short", "default", "long"]] = None
+    """
+    The preset to use for the model training duration. One of `debug`, `short`, `default`, `long`.
+
```
marielelandais, line 41:
```

    - `short` duration should last approximately 10 minutes.
    - `long` duration should last a few hours at most.

```

src/ansys/simai/core/data/geomai/models.py
marielelandais: Updates suggested inline.

image

src/ansys/simai/core/data/geomai/models.py
marielelandais: Duration will evolve in the following weeks. = Text to review in the future

src/ansys/simai/core/api/sse.py
```diff
@@ -144,6 +144,12 @@ def _handle_data_model_event(self, data):
                     logger.debug(
                         f"Unknown action {target['action']} of type {target['type']} received for job or resource event. Ignoring."
                     )
+        elif target["type"] == "geomai_model":
+            self.simai_client.geomai.models._handle_sse_event()
+        elif target["type"] == "geomai_prediction":
+            self.simai_client.geomai.predictions._handle_sse_event()
+        elif target["type"] == "geomai_training_data":
+            self.simai_client.geomai.training_data._handle_sse_event()
```
msd-11, line 152: Shouldn't it be 
```suggestion
        elif target["type"] == "geomai_model":
            self.simai_client.geomai.models._handle_sse_event(data)
        elif target["type"] == "geomai_prediction":
            self.simai_client.geomai.predictions._handle_sse_event(data)
        elif target["type"] == "geomai_training_data":
            self.simai_client.geomai.training_data._handle_sse_event(data)
``` 
?
Got errors : `Error handling SSE event: Directory._handle_sse_event() missing 1 required positional argument: 'data'. `

PR #159, opened 2025-06-11
feat: optional boundary conditions on optimization runs

msd-11:
Set boundary conditions to optional on `simai.optimizations.run_non_parametric()` and `simai.optimizations.run_parametric()` ~and add checks to validate boundary conditions by raising an exception when required boundary conditions are missing.~


msd-11:
Seems like the check is already done on user-api side (it also checks the value range).
I'll remove the check on SDK side and just set boundary conditions to be optional.

PR #171, opened 2025-07-02
fix: replace verify_gc_formula by process_gc_formula

u8slvn:
Replace old usage of `verify_gc_formula` by `process_gc_formula`.


awoimbee (approved):
Nice, let's make a patch release !

PR #182, opened 2025-09-02
chore: switch to niquest built in SSE support

tmpbeing:
Not a big fan of niquests so far, it's very poorly documented. Hopefully this fixes the issues with our tests.

### DIFF
--- a/src/ansys/simai/core/utils/sse_client.py
+++ b/src/ansys/simai/core/utils/sse_client.py
@@ -22,54 +22,62 @@
 
 import logging
 import time
-from typing import Callable, Optional
+from urllib.parse import urlparse
 
 import niquests
-import sseclient
-import urllib3
+import urllib3_future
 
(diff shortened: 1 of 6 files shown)


src/ansys/simai/core/utils/sse_client.py
```diff
@@ -22,62 +22,72 @@
...
                     yield event
                     self._last_event_id = event.id
             except (
                 StopIteration,
                 niquests.exceptions.ChunkedEncodingError,
                 niquests.RequestException,
                 EOFError,
-                urllib3.exceptions.MustRedialError,
+                urllib3_future.exceptions.MustRedialError,
+                # When the extension is closed the SSE
```
awoimbee, line 89: unfinished comment ?

awoimbee (approved):
Thanks, hope it won't cause too many problems !
Yeah seems like most python sync http clients see few contributions.

PR #189, opened 2025-09-17
feat: return more data in run_non_parametric optimization

msd-11:
`run_non_parametric()` optimization now returns `OptimizationResult` that contains more data that may be needed by the user : 
- The list of geometries created by the optimization
- The list of objectives (outcome value of the optimization)
- The `Optimization` object

### DIFF
--- a/src/ansys/simai/core/data/optimizations.py
+++ b/src/ansys/simai/core/data/optimizations.py
@@ -81,17 +81,9 @@ class OptimizationResult:
     ):
         self._optimization = optimization
         self._objectives: list[Dict] = objectives
-        self._geometries: list[Geometry] = []
-
-        workspace_geometries = self._optimization._client.geometries.list(
-            geometries[0]["workspace_id"]
-        )
-        workspace_geometries = {geom.id: geom for geom in workspace_geometries}
-
-        geometry_ids = [geom["id"] for geom in geometries]
-        for geom_id in geometry_ids:
-            if geom_id in workspace_geometries:
-                self._geometries.append(workspace_geometries[geom_id])
+        self._geometries: list[Geometry] = [
+            self._optimization._client.geometries._model_from(geom) for geom in geometries
+        ]
 
     @property
     def optimization(self) -> Optimization:


src/ansys/simai/core/data/optimizations.py
```diff
@@ -73,6 +73,40 @@ def _run_iteration(self, parameters: Dict) -> "_OptimizationTrialRun":
...
+    """Result for a non-parametric optimization."""
+
+    def __init__(
+        self, optimization: "Optimization", objectives: list[Dict], geometries: list[Dict]
+    ):
+        self._optimization = optimization
+        self._objectives: list[Dict] = objectives
+        self._geometries: list[Geometry] = []
+
+        workspace_geometries = self._optimization._client.geometries.list(
```
awoimbee, line 86: That doesn't scale well
And doesn't work with the fact that all our `list` methods should be replaced by `iter` (otherwise we end up with `training_data.list()` taking 5 minutes)

src/ansys/simai/core/data/optimizations.py
```diff
@@ -73,6 +73,40 @@ def _run_iteration(self, parameters: Dict) -> "_OptimizationTrialRun":
...
+            geometries[0]["workspace_id"]
+        )
+        workspace_geometries = {geom.id: geom for geom in workspace_geometries}
+
+        geometry_ids = [geom["id"] for geom in geometries]
```
awoimbee, line 91: Use an iterator, no need to build a list here
```suggestion
        geometry_ids = (geom["id"] for geom in geometries)
```

src/ansys/simai/core/data/optimizations.py
```diff
@@ -73,6 +73,40 @@ def _run_iteration(self, parameters: Dict) -> "_OptimizationTrialRun":
...
+        for geom_id in geometry_ids:
+            if geom_id in workspace_geometries:
+                self._geometries.append(workspace_geometries[geom_id])
```
awoimbee, line 94: This filtering logic must be documented, it's hard to understand why it's here

src/ansys/simai/core/data/optimizations.py
```diff
@@ -332,27 +366,33 @@ def run_non_parametric(
...
+                        trial_result["geometries"].append(trial_run.fields["geometry"])
                     else:
                         logger.debug("Trial run results did not match constraints. Skipping.")
                     logger.debug("Trial complete.")
                     progress_bar.update(1)
                 logger.debug("Optimization complete.")
                 progress_bar.set_description("Optimization complete.")
-            return iterations_results
+
+            optimization_result = OptimizationResult(
```
awoimbee, line 392: raw dicts are horrible to work with, I can't review this, trusting you on this one

awoimbee (approved):

src/ansys/simai/core/data/optimizations.py
msd-11: Well, I just saw that the trial runs return a `Geometry` object in the result, so I don't need to use `list` or `iter`, I updated this part of the code.

PR #185, opened 2025-09-05
fix: MustRedialError in SSE not being handled

tmpbeing:

PR #193, opened 2025-09-18
Replace requests/niquests by httpx

awoimbee:
* requests is ill maintained.
* niquests is causing issues:
  * https://github.com/jawah/urllib3.future/issues/266
  * #185
* the httpx API is way cleaner, resembles what I've seen in other languages

---

* The end to end tests are passing
* pytest-httpx is stricter than responses so the tests are better in the end
* the new SSE logic is much simpler and should work fine

The big user-facing difference is that httpx logs all requests at INFO level

https://github.com/user-attachments/assets/447879a3-af8f-4c79-8203-e6bbf27a4da0

### DIFF
--- a/tests/geomai/test_geomai_workspaces.py
+++ b/tests/geomai/test_geomai_workspaces.py
@@ -22,14 +22,11 @@
 
 from typing import TYPE_CHECKING
 
-import responses
-
 if TYPE_CHECKING:
     from ansys.simai.core.data.geomai.workspaces import GeomAIWorkspace
 
 
-@responses.activate
-def test_geomai_workspace_download_mer(simai_client):
+def test_geomai_workspace_download_mer(simai_client, httpx_mock):
     """WHEN downloading mer zip file
     THEN the content of the file matches the content of the response.
     """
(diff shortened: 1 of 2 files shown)


msd-11 (changes requested):

src/ansys/simai/core/api/sse.py
```diff
@@ -60,35 +63,70 @@ def __init__(self, config: ClientConfig, simai_client=None):
...
-        self.listener_thread = threading.Thread(target=self._sse_thread_loop, daemon=True)
-        self.listener_thread.start()
+        self._error_queue: queue.Queue[Exception] = queue.Queue()
+        self._flag_sse_started = Event()
+        logger.debug("Starting SSE listener thread.")
+        self._executor = ThreadPoolExecutor(
+            thread_name_prefix="SSEListener",
+            max_workers=1,
+        )
+        self._sse_listener: Future = self._executor.submit(self._sse_thread_loop)
```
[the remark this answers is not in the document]
awoimbee: Damn I thought they were :(
Fixed (had to add yet another try/except)

src/ansys/simai/core/api/mixin.py
```diff
@@ -199,25 +180,27 @@ def download_file(
...
-        response = self._request(request_method, download_url, **request_kwargs)
-        if monitor_callback is not None:
-            monitor_callback(int(response.headers.get("Content-Length", 0)))
-        logger.info("Starting download.")
+        logger.debug(f"Request stream {request_method} on {download_url}")
+        full_url = self.build_full_url_for_endpoint(download_url)
         try:
-            for chunk in response.iter_content(chunk_size=1024):
-                bytes_read_delta = output_file.write(chunk)
+            with httpx.stream(request_method, full_url, **request_kwargs) as response:
```
msd-11, line 189:
```suggestion
            with self._session.stream(request_method, full_url, **request_kwargs) as response:
```
Otherwise it will return HTTP code 401

msd-11 (changes requested):

src/ansys/simai/core/api/mixin.py
awoimbee: Woaw that's a nice catch ! Thanks

PR #198, opened 2025-09-24
improvements around authentication

awoimbee:
That's me trying to figure out the root cause of [sc-33966].

* I found that `def is_token_expired(self):` and `def is_refresh_token_expired(self):` both using `EXPIRATION_BUFFER` was weird as the access token being expired means "please refresh" while the refresh token being expired means "please reauthenticate"
* I added the log `"refresh token is expired"`
* I removed the useless scary log `INFO:ansys.simai.core.utils.auth:Could not read auth token from cache: [Errno 2] No such file or directory: '/home/arthur/.cache/ansys/pysimai/tokens-xyz123.json'`
* I found that `expires_in` and `refresh_expires_in` was cached, I don't think it was causing any issue but still, that was a bug waiting to happen

Note that long term maybe we should use [httpx-auth](https://pypi.org/project/httpx-auth/) anyways (after #193 is merged).


src/ansys/simai/core/utils/auth.py
```diff
@@ -46,15 +46,16 @@
 
 # polling can't be faster or auth server returns HTTP 400 Slow down
 DEVICE_AUTH_POLLING_INTERVAL = 5
+# Try to refresh tokens 300-400 secs before they go bad
+# - Randomized to prevent thundering herd
+# - Accounts for network latency and clock skew
+TOKEN_REFRESH_BUFFER = random.randrange(300, 400)  # noqa: S311
+TOKEN_EXPIRATION_BUFFER = random.randrange(5, 15)  # noqa: S311
```
msd-11, line 53:
```suggestion
TOKEN_REFRESH_BUFFER = random.randrange(300, 400)  # noqa: S311 # nosec
TOKEN_EXPIRATION_BUFFER = random.randrange(5, 15)  # noqa: S311 # nosec
```
For `check-vulnerabilities` job

msd-11 (approved):
Let's try this

src/ansys/simai/core/utils/auth.py
awoimbee: Nice, thanks !

PR #221, opened 2025-11-27
feat: predict scalars & rename boundary conditions

msd-11:
Rename `boundary conditions` to `scalars`  and add `PredictScalars` post processing with the possibility to use output scalars for building a model.

[sc-33563]
[sc-34397]


awoimbee (approved):
I mixed up `bc` and `gc`, sorry about that, LGTM !

PR #232, opened 2025-12-04
docs: format code blocks

msd-11:
Some code blocks were not rendered correctly in VS code.

[sc-33197]

### DIFF
--- a/src/ansys/simai/core/client.py
+++ b/src/ansys/simai/core/client.py
@@ -128,14 +128,15 @@ class SimAIClient:
             name: Name of the workspace that the client should switch to.
 
         Example:
-            .. code-block:: python
 
-                import ansys.simai.core as asc
+        .. code-block:: python
 
-                simai_client = asc.from_config(workspace="old plane")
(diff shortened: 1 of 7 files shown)


awoimbee (approved):

tmpbeing (changes requested):
This breaks formatting, example is no longer a section but written like simple text.

msd-11:
Arf, you are right. :grimacing: 
To not break the section we need to add a text between `Exemples:` and the beginning of `.. code-block:: python` (in order to have the formatting correctly in VS Code). Like:
```python
        Examples:
            Basic usage:

            .. code-block:: python

                import ansys.simai.core as asc
```

We should wait for @marielelandais 's return to see what phrasing to add, or should we go with `Basic usage`?

marielelandais (approved):
Looks good! Thanks @msd-11

msd-11:
Closing this PR, the error looks more on vscode side rather than our side since everything is rendered correctly in our documentation.

PR #234, opened 2025-12-05
chore: increase httpx default timeout

tmpbeing:
Some people are having read/write timeouts for unknown reasons (not just to our servers but also to s3 so likely spotty connection ?). This hopefully will help a bit scripts survive for users with spotty connections.


awoimbee (approved):
Thanks

PR #246, opened 2026-01-12
feat: add training data rename support

jeremiespiesser-extrality:
Add training data renaming support.

Awaiting API to be merged.

[sc-35630]


msd-11:
Don't forget to make the checks pass before merging

PR #258, opened 2026-02-04
Doc: add code example around automorphing

FlorianJactaAI:
Warning!!! This code is running with the current version and might be deprecated soon. We should change it so that it is working with next versions.

Also, I am not sure I can put the first code with the Pyvista rendering. Can you confirm? @marielelandais 

Resolves: https://app.shortcut.com/simai/story/36587/doc-create-code-example-around-automorphing

image

image

### DIFF
--- a/doc/source/examples/01_pysimai_ex/04-non_parametric_optimization.py
+++ b/doc/source/examples/01_pysimai_ex/04-non_parametric_optimization.py
@@ -111,7 +111,7 @@ ORGANIZATION_NAME = "<your_organization_name>"
 WORKSPACE_NAME = "<your_workspace_name>"
 CHOSEN_GEOMETRY_NAME = "<your_geometry_file.vtp>"
 
-# A bounding boxes must be defined as [xmin, xmax, ymin, ymax, zmin, zmax]
+# A bounding box must be defined as [xmin, xmax, ymin, ymax, zmin, zmax]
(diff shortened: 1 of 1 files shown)


marielelandais:
> Also, I am not sure I can put the first code with the Pyvista rendering. Can you confirm?

Before answering @FlorianJactaAI, I have questions: 
1/ Do we have an internal alternative?
2/ What if they upgrade the library and it doesn't work anymore. Is it possible? Who is going to verify it still works at each update?

marielelandais:
> > > Also, I am not sure I can put the first code with the Pyvista rendering. Can you confirm?
> > 
> > 
> > Before answering @FlorianJactaAI, I have questions: 1/ Do we have an internal alternative? 2/ What if they upgrade the library and it doesn't work anymore. Is it possible? Who is going to verify it still works at each update?
> 
> 1. We have no internal alternative in terms of rendering the boxes right now. The alternative would be to view this in Paraview and know the tool a bit.
> 2. Pyvista is used a lot by us but this special feature not too much (showing bounding boxes). I guess, it would be hard to know if the code still works. However, the functions shows are pretty basic and should hopefully not change so much.

It seems other pyansys libraries are using pyvista too

FlorianJactaAI:
> Nice ! Fix code style before merging please 🙏

Thank you! I will wait for the new release to synchronize with the possible new syntax

PR #273, opened 2026-03-09
feat: accessing objects created by the current user

msd-11:
The user can access `Project`, `Workspace`, `TrainingData`, `GeomAIProject`, `GeomAIWorkspace` and `GeomAITrainingData` that he created with the parameter `created_by_me` in `Object.list()` .

[sc-32881]


tmpbeing (approved):
I'm not opposed to merging that, it's solid work so I'll approve.

I do have a different idea though, we could have a module with a set of helpers/a class to build filters, one such helper/method would be to add the `created_by` filter with an easy way to put current user as the target.

There's ups and downs so I'm not saying my idea is better. 
On the plus side I think it's more maintainable and avoids "polluting" all the methods with that.
On the downside it's less discoverable and there's the issue of what if a specific method with filter doesn't actually support `created_by`. I haven't touched the filtering code so I'm not sure how well the errors are handled in the back, if we get a clean error message that might be okay.

PR #290, opened 2026-04-14
fix: define accepted file formats

msd-11:
Hard-code list of accepted file formats since most of the file formats were removed from the model manifest.


tmpbeing (approved):
Yikes. Why did they remove it if they're still accepted ? This whole feature is useless now.

msd-11:
The list should come from luggage from what I understand, but in the meantime, to avoid users being blocked to upload geometries, we hard-code it for now

PR #301, opened 2026-04-30
feat: GeomAI prediction auto-resolution

msd-11:
Add auto-resolution to GeomAI prediction, setting a resolution is not mandatory and the default value will be automatically selected.

Do not merge until user-api has not been merged.

[sc-38644]
[sc-38520]

### DIFF
--- a/doc/source/examples/02_generative_design_ex/02-generate_random_geometries.py
+++ b/doc/source/examples/02_generative_design_ex/02-generate_random_geometries.py
@@ -55,7 +55,6 @@ PROJECT_NAME = "new-bracket-project"  # Replace with your project name
 WORKSPACE_NAME = "new-bracket-project #1"  # Typically "{PROJECT_NAME} #{number}"
 OUTPUT_DIR = "random_geometries"  # Directory to save generated geometries
 NUM_GEOMETRIES = 5  # Number of random geometries to generate
-RESOLUTION = (100, 100, 100)  # Output resolution (x, y, z)
 
 ###############################################################################
 # Initialize the client and get the workspace
@@ -107,7 +106,6 @@ for i in range(NUM_GEOMETRIES):
     # Create prediction configuration
     config = GeomAIPredictionConfiguration(
         latent_params=latent_params,
-        resolution=RESOLUTION,
     )
 
     # Run the prediction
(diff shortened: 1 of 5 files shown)


msd-11:
@marielelandais @FlorianJactaAI I've updated the doc and user guide, if you find anything that needs to be updated, please add a comment :+1:

doc/source/examples/02_generative_design_ex/02-generate_random_geometries.py
```diff
@@ -146,7 +144,6 @@
 # -------------------------------------------
 #
 # - Latent parameters typically range from -3 to +3 for meaningful results.
-# - Adjust the resolution to balance quality and file size.
```
msd-11, line 149: Not sure if we need to remove this line @FlorianJactaAI

FlorianJactaAI:
This is nice! I think we want to still mention it somewhere. In the case of Stellantis, they had geometries that were created with decent resolution but because of how thin the geometry was, they had to go to 900.

> When working with complex or thin geometries, you might want to increase the resolution to get the best result.

-> something like this?

A side comment: Does the random sampling has been released? If so, I should also change the code so that we don"t do this random sampling I created.

doc/source/user_guide/generative_design_ug/generative_design.rst
```diff
@@ -11,18 +11,3 @@ You define this parameter to generate a geometry with a trained model.
 
 The number of floats must match the ``nb_latent_param`` your model was requested with.
 For more information, see Number of latent parameters.
-
-
-Resolution
```
marielelandais, line 16: I'm not sure we should have removed the entire section, though. I suggested changes for validation in Shortcut.

src/ansys/simai/core/data/geomai/predictions.py
```diff
@@ -49,17 +49,12 @@ class GeomAIPredictionConfiguration(BaseModel):
     Required.
     """
     resolution: Optional[Tuple[PositiveInt, PositiveInt, PositiveInt]] = None
-    """A list of three integers defining the number of voxels along the X, Y, and Z axes.
+    """Auto‑resolution is a system-driven feature that automatically selects the most appropriate reconstruction resolution at inference time based on training data.
```
marielelandais: Here as well, I suggested changes for validation in Shortcut.

marielelandais (commented):
Waiting for more info before being able to provide my feedback.

doc/source/examples/02_generative_design_ex/02-generate_random_geometries.py
FlorianJactaAI: You can comment the line and potentially say to use a higher value, explaining that increasing it improves the geometry resolution and helps capture finer details. However, this will also increase the prediction time.

tmpbeing (approved):

PR #303, opened 2026-05-07
Fix compatibility issue with new optimization object

FlorianJactaAI:
We have now an issue whether the non-parametric optim returns an Optimization object or an LegacyOptimizationResult. I am fixing the doc by having this "if" statement.

In this PR, I have also added a more robust code so that prediction issues are not an issue anymore. It also seems that Delta is not present in VTPs in the new release so this has been deleted too. And predictions are run after showing the graph as it was seen confusing for people to do the predictions here.

### DIFF
--- a/doc/source/examples/01_pysimai_ex/04-non_parametric_optimization.py
+++ b/doc/source/examples/01_pysimai_ex/04-non_parametric_optimization.py
@@ -202,7 +202,7 @@ optimization = simai.optimizations.run_non_parametric(
     detail_level=DETAIL_LEVEL,
     maximize=OBJECTIVE,  # Use 'minimize' parameter for minimization objectives
     show_progress=True,
-    part_morphing=PART_MORPHING,  # Comment to disable part morphing
+    # part_morphing=PART_MORPHING,  # Uncomment to enable part morphing
 )
 optimization.wait()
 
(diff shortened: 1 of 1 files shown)


doc/source/examples/01_pysimai_ex/04-non_parametric_optimization.py
```diff
@@ -195,63 +196,37 @@
...
-# You can get back your optimization from its ID when done asynchronously.
-# Run predictions on all generated geometries to evaluate their performance:
-
-predictions: list[Prediction] = []
 
-for geom in optimization.list_geometries():
-    print(f"Running prediction for: {geom.name} (ID: {geom.id})")
-    prediction = geom.run_prediction(scalars={})  # Scalars must match your workspace configuration
-    predictions.append(prediction)
+from ansys.simai.core.data.optimizations import Optimization
```
awoimbee: import not at top of file ?

doc/source/examples/01_pysimai_ex/04-non_parametric_optimization.py
```diff
@@ -195,63 +196,37 @@
...
-VTPS_FOLDER = f"{OUTPUT_FOLDER}/optimization_{optimization_id}/VTPs"
-os.makedirs(VTPS_FOLDER, exist_ok=True)
-
-for prediction in predictions:
-    # Wait for prediction to complete
-    prediction.wait()
-
-    geom = prediction.geometry
-    geom_name = geom.name.split(".")[0]
+optimization_id = (
```
awoimbee, line 216: Maybe explain this trick

awoimbee (approved):

PR #310, opened 2026-05-21
feat: add optimization field to optimization for backwards compatibil…

jeremiespiesser-extrality:
[sc-38951]

Adds a `.optimization` field to class Optimization for backwards compatibility.
Removes the need to do the `if` clause in
```
optimization.id if isinstance(optimization, Optimization) else optimization.optimization.id
```

### DIFF
--- a/doc/source/examples/01_pysimai_ex/04-non_parametric_optimization.py
+++ b/doc/source/examples/01_pysimai_ex/04-non_parametric_optimization.py
@@ -117,7 +117,6 @@ plotter.show()
 import os
 
 import ansys.simai.core as asc
-from ansys.simai.core.data.optimizations import Optimization
 from ansys.simai.core.data.predictions import Prediction
 
 ORGANIZATION_NAME = "<your_organization_name>"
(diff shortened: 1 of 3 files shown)


msd-11 (approved):
Is `LegacyOptimization` something that is going to disappear in the future ?

jeremiespiesser-extrality:
Yes It will once we switch to only server side optims

doc/source/examples/01_pysimai_ex/04-non_parametric_optimization.py
```diff
@@ -213,9 +212,7 @@
 # Get the optimization ID for either a server-side or legacy optimization,
 # and print the generated geometries and objective values:
 
-optimization_id = (
-    optimization.id if isinstance(optimization, Optimization) else optimization.optimization.id
-)
+optimization_id = optimization.optimization.id
```
tmpbeing: Was the previous method broken ? I don't think we should recommend users to use that approach when we probably don't want to keep it long term.

src/ansys/simai/core/data/optimizations.py
```diff
@@ -73,6 +74,11 @@ def __init__(self, *args, **kwargs):
         self._geometries = None
         self._objectives = None
 
+    @property
+    def optimization(self) -> Self:
+        """Returns self. Present for backwards compatibility with LegacyOptimizationResult."""
```
tmpbeing: I'd add a deprecation warning saying this will be removed when the legacy client-side optimization will be removed.

PR #317, opened 2026-06-15
fix: refresh token with same client id

jeremiespiesser-extrality:
if refresh token has azp still in legacy "sdk", request a new token with the same "sdk" client_id (else the backend will reject it)

Doesn't touch new token requests which will request "com.ansys.simai.sdk".

[sc-new-story]


awoimbee (approved):
You should mark this as a temporary fix to make legacy tokens work.

We could instead update `_auth_hash` to include the client id.

u8slvn (approved):

PR #319, opened 2026-06-16
fix: use __args__ for isinstance checks for Python 3.9 compatibility

msd-11:
That should fix the failing test on python 3.9

### DIFF
--- a/src/ansys/simai/core/data/geomai/workspaces.py
+++ b/src/ansys/simai/core/data/geomai/workspaces.py
@@ -126,7 +126,7 @@ class GeomAIWorkspace(DataModel):
         """
         data = self._client._api.download_geomai_workspace_latent_parameters(self.id, file)
         if file:
-            if isinstance(file, File) and not isinstance(file, (str, Path)):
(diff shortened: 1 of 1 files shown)


src/ansys/simai/core/data/geomai/workspaces.py
```diff
@@ -126,7 +126,7 @@ def get_latent_parameters(
         """
         data = self._client._api.download_geomai_workspace_latent_parameters(self.id, file)
         if file:
-            if isinstance(file, File) and not isinstance(file, (str, Path)):
+            if isinstance(file, File.__args__) and not isinstance(file, Path.__args__):
```
awoimbee, line 129: Why remove `str` from the type check ? What's the impact ?

msd-11: `Path` is defined as `Union[pathlib.Path, str, os.PathLike]`, which includes `str`, the `str` was redundant. It should have no impact.

awoimbee (approved):

PR #331, opened 2026-07-08
fix: for server side optim, request a token matching the coreml versi…

jeremiespiesser-extrality:
For a server side optim, if coreml_version belongs to  [3.16.0, 3.18.0 [, the used offline_token must be of client_id == "sdk", and if coreml_version is over or equal to 3.18.0, the token's client id must be "com.ansys.simai.sdk". **Update** : 3.20.0 and upwards can support both.

This PR aims to, if it detects a mismatch client_id between the provided offline_token and what the model expects, print a warning and ask the server for a new token matching what the model expects.

Cases like this should be quite rare.

### DIFF
--- a/src/ansys/simai/core/data/optimizations.py
+++ b/src/ansys/simai/core/data/optimizations.py
@@ -24,9 +24,16 @@ import logging
 import warnings
 from dataclasses import dataclass
 from typing import Any, Dict, List, Literal, Optional, Union
+import re
 
 from tqdm import tqdm
+from ansys.simai.core.utils.auth import (
+    _get_offline_token_private,
+    OIDC_CLIENT_ID,
+    _decode_authorized_party,
+)
 from wakepy import keep
+from itertools import chain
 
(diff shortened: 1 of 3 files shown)


src/ansys/simai/core/data/optimizations.py
```diff
@@ -790,3 +823,31 @@ def _build_objective(minimize: list[str], maximize: list[str]) -> dict:
...
+            elif minor <= 19:
+                return OIDC_CLIENT_ID
+            # from 3.20 onwards : both are supported
+            else:
+                return None
+        if major > 3:
+            # TODO : update when coreml can start supporting both
+            return None
+    else:
+        return None
```
awoimbee, line 852: :cry-blood:

Can't we just disable optim on these 2 old versions ? And forget that the client_id was ever `sdk` ?

tmpbeing (approved):
If it's better for users, we can have a bit of "ugly" code. Thanks !

PR #351, opened 2026-07-23
chore: update timeout value and chunk size for file upload

msd-11:
Increase timeout value and decrease chunk size for file upload. This should fix timeout issues with slow internet and big files.

[sc-39769]


awoimbee (approved):
Thanks.
It's clearly still not perfect but much better

PR #352, opened 2026-07-23
chore: deprecate optimization_id and use id in get optim method

jeremiespiesser-extrality:
Reason : consistency with other get methods in the sdk

Not a breaking change but a deprecation (marked optimization_id as deprecated in docs)


msd-11 (changes requested):
Please add a `warnings.warn()` when `optimization_id` is used and add what parameter to use instead (`id`) in the `warnings.warn()`

msd-11 (approved):
Thx (fix code style before merging 🙏 )

u8slvn (approved):
LGTM

PR #358, opened 2026-07-24
Fix PySimAI examples and make them faster

FlorianJactaAI:
Related to this: https://app.shortcut.com/simai/story/40001/fix-pysimai-examples-on-upload-data

The goal is to fix an issue on the upload and make this code a bit better. Tigerteam users could wait a long time with this `simai.training_data.list()`.


msd-11 (commented):
If `training_data.list()` is too slow, either use `training_data.iter()` or, in your case, use filters `training_data.list(filters=...)` you have operators like `IN` that should be useful in this use case.

FlorianJactaAI:
> If `training_data.list()` is too slow, either use `training_data.iter()` or, in your case, use filters `training_data.list(filters=...)` you have operators like `IN` that should be useful in this use case.

I am using the "get" method with the name. It is using a similar pattern beneath?

FlorianJactaAI:
@msd-11 , can we merge it? Some people have complained about the examples

msd-11 (approved):
Ok for me

The filters in `training_data.list()` allow you to retrieve all the TDs that match the filters within a single call instead of doing one `get()` call for each TD

FlorianJactaAI:
I don't really get how you do that in practice 🤔

PR #365, opened 2026-07-30
docs: Put the explainer in a box to better organize the section

alth-ansys:
In this PR, I suggest to refactor the Resolution section in Generative Design doc to isolate the "how does auto resolution work" 

Before:
image

After:
image


msd-11 (approved):
Thanks !

PR #393, opened 2026-09-11
migrate httpx to httpx2

awoimbee:
httpx is dead, let's use pydantic's fork: httpx2

### DIFF
--- a/src/ansys/simai/core/utils/transport.py
+++ b/src/ansys/simai/core/utils/transport.py
@@ -1,0 +1,131 @@
+# Copyright (C) 2023 - 2026 ANSYS, Inc. and/or its affiliates.
+# SPDX-License-Identifier: MIT
+#
+#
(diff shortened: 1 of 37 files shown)


src/ansys/simai/core/utils/transport.py
```diff
@@ -0,0 +1,131 @@
...
+class Retry:
+    """Configuration for :class:`RetryTransport`."""
+
+    def __init__(
+        self,
+        total: int = 10,
+        allowed_methods: Optional[Iterable[str]] = None,
+        status_forcelist: Optional[Iterable[int]] = None,
+        retry_on_exceptions: Optional[Iterable[Type[Exception]]] = None,
+        backoff_factor: float = 0.0,
```
msd-11: Why not setting the default to `0.5` or `1` ? When `service is unavailable` or `too many request` we want perhaps to wait some time before retrying that increases after each request.

PR #394, opened 2026-09-14
chore: download_latent_parameters_json now raises an error as part of depreciation process

tmpbeing:


msd-11 (commented):
PyAnsys best practices suggest to use `AttributeError` or to make a custom error of deprecation https://dev.docs.pyansys.com/coding-style/deprecation.html
Since we use `PySimAIDeprecationWarning` perhaps we should use something similar for deprecation errors ?

msd-11 (approved):
Thanks

